PR #38, opened 2020-02-05
Feature: Adjusted and New Components

bspeare:
The changes in this PR are:

- Updated vertical video component with new videos, content, and layout
- Updated BG Image component with new format
- Updated Card component with new design
- Added new CTA block to Blocks folder
- Added new layout in Templates directory for home page
- Moved components in 03-global that use render tag into preview folder (including search-toggle--btn.twig)
- Added new footer row call-to-action links above footer
- Adjusted Hero component styles


joewhitsitt:
just a quick note on the the stat `display: inline-table`. the cases are few, but there is potential because of the prefix/suffix spans that a stat would break up into lines. experience from D7, if the font doesn't scale down, people put big stats in e.g. $1,000,000 and it breaks the $ from the amount when columns get smaller.

PR #47, opened 2020-02-18
Feature: Updates to footer templates and scss

pyrello:
* Updated `gulp watch` command to only trigger 'fractal start' at the beginning. 
* Removed unintentionally nested links from menus in footer config context. 
* Added new generic templates for links, menus, and menu items.
* Removed unnecessary footer-specific templates for the above items. 
* Added socket menu template. 
* Re-factored footer template to use new templates.
* Removed variants from footer config for removed templates.
* Fixed an if condition nesting problem in the footer template. 
* Re-factored footer scss file to apply the same style to all levels of footer links.


bspeare:
@pyrello, this looks good but the button row above the footer is now broken.  You/I will need to fix that before we merge this into master.

PR #72, opened 2020-03-12
Feature: Internal navigation updates

pyrello:


bspeare:
@pyrello It looks like the `@menu` footer links are broken since the addition of the item.config.yml and menu.config.yml.

pyrello:
@bspeare Should be fixed now.

Issue #71, opened 2020-03-11
Documentation: Creating a new uids release


bspeare:
First, in uiowa/uids on master

1. `npm version --help` to view options
2. `npm version patch -m "https://github.com/uiowa/uids/compare/v1.4.1...v1.4.2"`
3. `git push`
4. `git push --tags`
5. Go to https://github.com/uiowa/uids/tags and click on the "..." button to the right of the release Place the release number in the title field without the v -> 1.4.2.
6. Paste in the compare link into the description area `https://github.com/uiowa/uids/compare/v1.4.1...v1.4.2` 
7. Then click "create release" button

Next go to uiowa/uiowa and create a feature branch

1. `git checkout -b feature_uids_bump`
2. change uids version number in uids_base package.json 
3. Then run `yarn upgrade @uiowa/uids`
4. `git commit -am "Bumped up uids version to 1.4.2"`
5. `git push -u origin feature_uids_bump`
6. Create pull request for feature branch

PR #73, opened 2020-03-17
Move fractal to devDependencies.

### LINKED ISSUE #45: Move fractal to dev dependencies
So downstream projects don't have to depend on it also. Fractal is a needed to run the UIDS but not necessarily needed for applications to use the UIDS.

richardbporter:
Resolves #45 

The thought here is that Fractal is not necessarily needed to consume design components whereas Gulp is considering no compiled CSS is provided. Applications can still install devDependencies and leverage Fractal if desired.


pyrello:
@richardbporter What is needed to move this PR forward? I think the concept is sound and it will help with build time.

richardbporter:
Here is what I think needs to happen:

- Create a corresponding PR for the now deprecated HDS.
- Move custom theme package.json devDependencies to dependencies.
- Make `yarn install` and Gulp tasks environment-aware so dev dependencies are not installed.

The last step probably relates to https://github.com/uiowa/uiowa/issues/1289. I'm a little fuzzing making that happen.

https://classic.yarnpkg.com/en/docs/cli/install/#toc-yarn-install-production-true-false

https://travis.media/how-to-run-gulp-tasks-according-to-environment-production-or-development/

richardbporter:
I'm going to merge this and then work through the release process outlined in #71 to refresh my memory. Then I can hopefully address #100 as well.

I tested the UIDS locally and everything looked OK. 👍

PR #96, opened 2020-04-03
Accordion uids

GaryRidgway:


bspeare:
It looks and works great! I'm not sure why that CSS is compiling into the dist folder.

bspeare:
@GaryRidgway Maybe add some `focus` and `hover` styling on the accordion button?  Perhaps add underline and some styling to the arrow might be good.

GaryRidgway:
I have fixed the CSS compiling in to dist, added the svg fallback, removed the need for unique IDs per accordion, and generated accordion item IDs programmatically

bspeare:
It looks great! I updated some padding and font sizes.  Thank you!

components/01 - elements/accordions/accordions.js
```diff
@@ -0,0 +1,34 @@
+// This would be a unique ID for every accordion.
```
pyrello, line 1: Is this comment necessary, since the line below is selecting by class now?

components/01 - elements/accordions/accordions.twig
```diff
@@ -0,0 +1,29 @@
...
+      >
+        {{ item.title }}
+        <i aria-hidden="true" focusable="false" class="fas fa-chevron-up"></i>
+      </button>
+    </h2>
+    {% if accordion.multiselectable == true %}
+      {% set accordion_type = 'multi' %}
+    {% else %}
+      {% set accordion_type = 'single' %}
+    {% endif %}
```
pyrello, line 23: Can this be set at the beginning of the template file?

GaryRidgway: Yes, I have done this in my latest round of commits, good suggestion!

components/01 - elements/accordions/accordions.js
GaryRidgway: Thanks for the catch, this has been removed.

PR #105, opened 2020-04-06
Feature: Card block text area enhancements

bspeare:
In this PR:

Updated card.scss to allow for inline links in card text area since SiteNow allows for this in the card component.  Also added a conditional on the `bttn` to display only if `card_link_title` exists.  This PR also includes a fix for #7. 

# How to test 

1. View http://localhost:3000/components/detail/card-block--default and try to use the link within the text area of the first card. 
2.  Edit the `yml` file for the card and remove the `card_link_url` or the `card_link_title` and verify that the button no longer appears below the card. 
3.  Edit the `yml` file for the card and remove the `card_title` and verify that the title no longer appears but the button does appear using an `a` tag and appears full-width inside the card and is clickable for the entire card area.


pyrello:
@bspeare It looks like you may need to check your settings with regard to code style. It looks like twig files are formatting with a 4 space indent, and I think it should be 2.

bspeare:
@pyrello Thanks for pointing that out.  It looks like I was having this issue: https://github.com/github/VisualStudio/issues/1524.

Issue #148, opened 2020-04-23
As an end user, I can view alerts in colors that reflect universal norms.


bspeare:
- [ ] Update Alerts component to reflect a more universal color palette
- [ ] Update uids_base to reflect these updates

# Additional notes:

https://designsystem.digital.gov/components/alert/

- blue 
- yellow
- red 
- green

PR #151, opened 2020-04-24
Feature: uids alerts updates

bspeare:
This PR addresses issue #148 

# How to test

Look through http://localhost:3000/components/detail/alerts--default and let me know what you think about the color and style adjustments for the alerts. 

The alerts still incorporate the brand colors, but using them to appear more like a standard alert palette.


bspeare:
@richardbporter It shouldn't require any changes for now.  If it does, I'll open a separate PR so that we can work through those issues if there are any.

PR #193, opened 2020-05-19
Auditing UIDS to prepare for release

pyrello:
Checklist of items is at #206


pyrello:
Created #206 to take over for the list on this PR.

pyrello:
So, I ended up doing some pretty significant refactoring on the whole project. I think its structured a little bit better for moving forward with being a shared resource. Also, I was able to figure out how to get some of the basics of theming working:

Screen Shot 2020-05-23 at 10 18 03 PM

It is important to know that things have moved around quite a bit and I have changed some things about the build process.

Here are some of the major highlights:
* The top-level `assets`, `components`, and `documentation` folders have been moved to a `src` directory.
* The `dist` folder isn't currently being generated, which will need to be fixed. This process needs to be re-thought so that we are only published items to `dist` that we want people to use. At this point it should be just css and javascript, I think. Possibly the scss files as well.
* When running `npm run watch`, the temporary version of the site that is served is placed in a `tmp` folder, which is in the `.gitignore` file, so it is not actually a part of the project.
* When running `npm run build`, the files get placed in a `www` folder. As part of the deployment process, this command is triggered and then the contents of the `www` folder are moved into the root of the project. This is what gets deployed to Github pages.
* The `config.json` file has been removed, per @richardbporter's [comment](https://github.com/uiowa/uiowa/issues/1402#issuecomment-630404547).
* The `fractal.js` file has been renamed to `fractal.config.js` command line messages after moving to version 1.3.0.
* I have removed all the dependencies from the `package.json` file that were not being used.
* In the process of refactoring `gulpfile.js`, I switched some dependencies and added some other dependencies.
* I have left a version of the root-level `components` directory in place. This should help ensure that we are not immediately breaking compatibility with uids_base. As we switch uids_base over to using the components at their new location, we can remove them from this folder and eventually remove it all together. The project is configured to read components from `src/components` directory now, so the items in the root-level directory do not get rendered.
* There is probably something else I am forgetting.

The checklist in #206 may not be totally up-to-date, but is still a good guide for what still needs to be done. Moving forward, components should be copied from `components` into `src/components` and additional modifications should be made in the new copy.

pyrello:
FYI, here is the repo that I have been referencing when making updates: https://github.com/24ways/frontend

bspeare:
@pyrello What's the best way to override the theme.css?  In particular, I'd like to override:

```
h1 {
    font-feature-settings: "dlig" 1;
}
```

bspeare:
@pyrello I have hidden the components that I wanted to hide and made sure that the colors used in the new fractal theme are accessible.  As long as the legacy `components` remain the same for putting out bug fixes this week I think this can go live if you think it is ready.

PR #195, opened 2020-05-19
Form styles

GaryRidgway:


bspeare:
@GaryRidgway Is there a way to style the upload input?

Just some example ideas:
https://css-tricks.com/snippets/css/custom-file-input-styling-webkitblink/
https://medium.com/@preetishhs24/styling-html-file-input-button-with-pure-css-without-losing-the-file-name-e1c344e5da43

bspeare:
@GaryRidgway The spacing looks better if this is set to `0`: https://github.com/uiowa/uids/blob/eafdefea5acbfed308317c12ed88f70a7580e25b/components/01%20-%20elements/forms/_forms.scss#L19.   Does this break anything if this is changed?

GaryRidgway:
> @GaryRidgway The spacing looks better if this is set to `0`:
> 
> https://github.com/uiowa/uids/blob/eafdefea5acbfed308317c12ed88f70a7580e25b/components/01%20-%20elements/forms/_forms.scss#L19
> 
> . Does this break anything if this is changed?

Sure! i don't think i actually changed that,  just found an extra legend declaration and was cleaning things up with it moving around. 0 is fine!

bspeare:
@GaryRidgway Is there way to add the missing images to UIDS and use an image path variable to set the image for UIDS and uids_base?  https://github.com/uiowa/uids/blob/e587d6ea0f160e8efc2b18b50d6ef573f8c7ca9f/assets/scss/_utilities.scss#L727

GaryRidgway:
> @GaryRidgway Is there a way to style the upload input?
> 
> Just some example ideas:
> https://css-tricks.com/snippets/css/custom-file-input-styling-webkitblink/
> https://medium.com/@preetishhs24/styling-html-file-input-button-with-pure-css-without-losing-the-file-name-e1c344e5da43

These are very different from what i had seen as options, but i can give it a go! it is super weird because the html for the button and the label looks like this:


`<input data-drupal-selector="edit-file-upload-upload" type="file" id="edit-file-upload-upload" name="files[file_upload]" size="22" class="js-form-file form-file">`

GaryRidgway:
> @GaryRidgway Is there way to add the missing images to UIDS and use an image path variable to set the image for UIDS and uids_base?
> 
> https://github.com/uiowa/uids/blob/e587d6ea0f160e8efc2b18b50d6ef573f8c7ca9f/assets/scss/_utilities.scss#L727

I'll give this a go but I will try and prioritize the other things first i think

bspeare:
@GaryRidgway for the inputs, it looks like even adding something like this would make it look better:

```
margin: 0;
font-family: inherit;
font-size: inherit;
line-height: inherit;
```

GaryRidgway:
upon further review i have a few things to do still, and i need to finish a multi file upload option.

GaryRidgway:
also row weights dont seem to be working for the multi add text fields???

PR #204, opened 2020-05-22
update with finalized wordmark 

joewhitsitt:
relates to https://github.com/uiowa/uiowa/issues/1322

PR #211, opened 2020-05-26
finalized wordmark update

joewhitsitt:
Re-attempt of #204 

Got help from Sondra on this one.


bspeare:
@joewhitsitt This looks a lot better!    @pyrello What do you think?

PR #220, opened 2020-05-28
Feature: Header adjustments / prefers motion update

bspeare:
In this PR:

- Logo tab width update
- Header.js update to address https://github.com/uiowa/uiowa/issues/1665
- Refactor header.scss and iowa-bar.scss to separate styles in preparation for https://github.com/uiowa/uids/issues/222
- Optimized brain pattern svg files
- Set `prefers-reduced-motion` check in hero.js for https://github.com/uiowa/uiowa/issues/1680.


src/assets/scss/iowa-bar-dist.scss
```diff
@@ -0,0 +1,19 @@
+@import url('https://use.fontawesome.com/releases/v5.8.2/css/all.css');
```
pyrello, line 1: I can't think of anything that we need this for.

PR #241, opened 2020-06-10
tweak a11y related to search-toggle

joewhitsitt:
This PR attempts to resolve the a11y issue reported by @inspector508 along with some other tweaks:

>Search button doesn’t announce changes in aria-expanded value.

https://github.com/uiowa/uiowa/issues/1437#issuecomment-634686264

The initial problem appears to be that the current implementation forces a focus change to the search input before the aria-expanded value on the toggle button is announced. The force focus change was done to lead the user back up the DOM where the search input is instead of out to the next element (main menu).

I changed the DOM order after talking with @bspeare and removed the forced focus change. I also added similar aria-controls information to the markup similar to the accordion component. I also believe there was misuse of `aria-expanded` on the search-overlay element where aria-hidden makes more sense. The css targeting the aria-expanded attribute had to be changed.

To Test:
- Pull branch, build, start.
- Go to http://localhost:3000/components/preview/header-old--default and try out the updated behavior
- Check to make sure you can still click out of the the search area to close the search area.

Below is a screencast demonstrating current behavior followed by the change. Notice on the current behavior the button expanded is not announced and focus immediately switches to the search input. The updated example announces the expanded change and allows natural tab navigation to the search input.

[OneDrive video link](https://iowa-my.sharepoint.com/:v:/g/personal/jwhitsit_uiowa_edu/EWbPXa--SmJGr_ZEN0pS5KABFjyIS7HtSieXPwmBW95QRg?e=2Ucd9a)

### DIFF
--- a/src/components/search-toggle/_search-toggle.scss
+++ b/src/components/search-toggle/_search-toggle.scss
@@ -102,7 +102,7 @@
     }
   }
 
-  &[aria-expanded="true"] {
+  &[aria-hidden="false"] {
     background: $white;
 
     @include breakpoint(menu) {
@@ -347,7 +347,7 @@
 
 
 
-  &[aria-expanded="true"] {
+  &[aria-hidden="false"] {
     z-index: 1;
     visibility: visible;
     transition: opacity .3s;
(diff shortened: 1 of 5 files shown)


bspeare:
@joewhitsitt I changed one element back to `aria-expanded=true` in the scss for the toggle close button.  Does that mess anything up with your changes?

joewhitsitt:
Nope. sorry, I went search/replace crazy.

PR #250, opened 2020-06-22
Feature: Banner slider prototype

bspeare:


pyrello:
@joewhitsitt Would love to have an a11y review, if you have the time!

joewhitsitt:
Looks like some things will be solved by content (e.g. Read more link, alt text) but here are a few notes. Careful what you ask for:

- Doesn't appear to be keyboard accessible (arrow keys or tab+enter does not change the slide). With inspector open (chrome) my tabbing does seem to have an effect for a moment, but then the animation rolls back to the initial slide.
- The tab focus seems to be cut off and at least visually it isn't clear what I am focusing on.
- Don't use the <i> for font awesome icons. Use <span> instead.
- Some more aria markup similar to a carousel would be nice to sprinkle in but I understand that this is the non-drupal version.

I'd be willing to do a voice over test after some of these have been addressed.

Personal opinion: the transition time is faster than what I am used to and a bit surprising if I am not expecting it.

pyrello:
Made the following updates:
* Switch font awesome icon to use `<span>` instead of `<i>` - If this is a common accessibility best practice, we should maybe think about making an icon component and implementing that so its consistent everywhere.
* Implemented left and right arrow movement through slides.
* Removed the `onfocusout` and `onmouseout` events from the slides so that they stay on whatever slide was last encountered.

Some more things that still need to be done:
* Adjust javascript so that it can support multiple sliders on the same page - currently, it will just grab all slides on the page and they'll all be grouped together in terms of interaction.
* Compare javascript against Accordion implementation to see if there are any obvious similarities and optimizations that can be made.
* Do we want to make banners use buttons for links across the board? Or do we want to leave the possibility of using a non-button link?

bspeare:
I think it could be useful to have the banner function like the card somewhat.  Display a link button if there is link text provided and if not, link up the entire card if there is a link url.

pyrello:
I'm for that. We can generalize that functionality to not be card-specific

pyrello:
I re-wrote the javascript to use an OO-style implementation of the slider. This allowed for more easy scoping of events to the individual sliders. I also added a variant example with multiple sliders for testing.

joewhitsitt:
- Thanks for making the i -> span change. We have fontawesome (admin/config/content/fontawesome) settings switched to span because siteimprove flags it and <i> is deprecated html tag. I had not previously thought about how the uids was implementing them.
- I was still encountering some odd behavior with the keyboard, but Chrome inspector and VoiceOver seem to fix it.
- VoiceOver told me enough information about the slides I was interacting with. Not sure if JAWS will be as happy without more parent/child labeling.

I think you can move forward so that more eyes can evaluate it.

pyrello:
As a side note, I thought that `<i>` was no longer deprecated in HTML5.

joewhitsitt:
You are right. Ben made a similar point a while back. It is still carved in my mind though.

`span` is more semantic and siteimprove (new law of the land) dings a11y scores on `i`

PR #271, opened 2020-07-08
increase a11y for current footer links

joewhitsitt:
Screen Shot 2020-07-08 at 12 57 21 PM
Related to the design meeting/#uids conversation. Short term solution to improve siteimprove score on uiowa.edu until styles can be expanded to cover more types of content in the footer.


bspeare:
Here's the live preview url: https://uids.brand.uiowa.edu/branches/footer-link-a11y//components/detail/footer.html

PR #281, opened 2020-07-13
Fix "Redundant WAI-ARIA attribute" on search button

joewhitsitt:
uiowa/uiowa/issues/1119


joewhitsitt:
Actually, marking as blocked. Doing a lot of reading and until IE11 support goes away, I am not sure this should be done.

joewhitsitt:
Closing for now.

Issue #284, opened 2020-07-17
Adjust events component row alignment


bspeare:
# Problem

`align-items: center;` creates vertical centering on the `.events-card__row`. 

<img width="1580" alt="Screen Shot 2020-07-17 at 4 30 21 PM" src="https://user-images.githubusercontent.com/1036433/87832162-f1e9fc00-c84a-11ea-8dee-eeb2192a68e6.png">

# Proposed Solution

Remove vertical centering and allow it to align to the top.

<img width="1530" alt="Screen Shot 2020-07-17 at 4 28 11 PM" src="https://user-images.githubusercontent.com/1036433/87832173-fb736400-c84a-11ea-910a-2aa462cb56c4.png">

This will help in other areas like in this type of row:

<img width="1325" alt="Screen Shot 2020-07-17 at 4 32 57 PM" src="https://user-images.githubusercontent.com/1036433/87832266-3c6b7880-c84b-11ea-954f-95e6bd74b455.png">

PR #285, opened 2020-07-17
Feature: Event component centering adjustment

bspeare:
Resolves https://github.com/uiowa/uids/issues/284

https://uids.brand.uiowa.edu/branches/feature_event_alignment_fix/components/detail/events-block.html

### DIFF
--- a/src/components/event/event.config.yml
+++ b/src/components/event/event.config.yml
@@ -28,7 +28,7 @@ context:
         events_card_title: College of Education 9 a.m. Commencement Ceremony
         events_card_link_title: Read Article
         events_card_link_url: https://uiowa.edu/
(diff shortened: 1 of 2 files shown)


joewhitsitt:
Noticed that the images are resized based on the width of the event text. Is that by design?
Screen Shot 2020-07-20 at 12 19 17 PM

bspeare:
@joewhitsitt Good observation!  I'd like to fix that.

bspeare:
@joewhitsitt there's no rush at all, but if you have a second can you confirm that the images sizes remain the same and not based on text width: https://uids.brand.uiowa.edu/branches/feature_event_alignment_fix/components/preview/events-block.html.

PR #294, opened 2020-07-31
Updates to dist.

pyrello:


pyrello:
Closing this PR as it is very out-of-date and probably not worth pursuing anymore. All of this will be handled differently in the 4.x branch since we are no longer using gulp.

PR #327, opened 2020-10-08
Event Card Variant

GaryRidgway:
This PR is to create a variant for the card component that translate everything that the event component could do in to it.


bspeare:
Should we update the class names in the card details?  The author class seems like it should change to `subtitle` or something related. 

- `card__date`
- `card__author`
- `card__label`

pyrello:
I was wondering about this too. Specifically, for the date, I was wondering about whether we might ever want to include other types of icons in a similar fashion for other arbitrary pieces of data that get included in the card. Since we use it for everything :)

GaryRidgway:
author has been replaced! We are subtitle now!

pyrello (approved):
Tested via integration in uiowa/uiowa project.

Issue #400, opened 2020-11-02
Update card template to use background.scss component for a11y color contrast


bspeare:
# Problem

Once https://github.com/uiowa/uids/pull/327, is merged, the card component will need to be tested on the kitchen sink test page to make sure that the color contrast of the headline, text, and bttn are all accessible using all background color options. 

# Proposed Solution

Implement `background.scss` on the card component and test color contrast issues on the kitchen sink page and adjust styles as necessary.

PR #401, opened 2020-11-02
Feature: card background colors a11y

bspeare:
Resolves https://github.com/uiowa/uids/issues/400.


src/components/card/card.scss
```diff
@@ -464,3 +483,60 @@
...
+}
+
+.card {
+
+  &.bg--white,
+  &.bg-pattern--brain,
+  &.bg--gold .bg--white,
+  &.bg--gold .bg-pattern--brain,
+  &.bg-pattern--brain--gold .bg--white,
+  &.bg-pattern--brain--gold .bg-pattern--brain,
```
pyrello, line 531: I don't understand the use of this unless you think it is possible that we are going to allow background colors inside of cards - in addition to the card having a background color itself.

pyrello (approved):
Appears to work. Nice work!

PR #409, opened 2020-11-10
Feature: Hotfix added important breakpoints for card padding

bspeare:


GaryRidgway (approved):
Minimally reviewed, but this seems super reasonable.

Issue #394, opened 2020-10-27
Update banner component to allow for a background video


bspeare:
- [x] move over hero.js and refactor 
- [x] Add pause/play button styles to banner. 
- [x] Update template refactor
- [x] More scss updates

[A PR to track progress for this has been started here](https://github.com/uiowa/uids/pull/412)

joewhitsitt:
Will need to account for what the expected behavior is when different banner heights are selected. I remember this being a concern with the video team when we first built the hero display. (e.g. cropping/scaling/center focus, etc.)

bspeare:
Closed via https://github.com/uiowa/uids/pull/412.

Issue #393, opened 2020-10-27
Update banner block styles to allow for a button row and update banner.js


bspeare:
- [x] Change banner.js if there is more than one link
- [x] Display multiple links in a row
- [x] Allow button to right click and open link in new window
- [x] Allow anchor links to be used within the inline text and buttons

5 points

bspeare:
We will need something like this:
```
.bttn--row .bttn + .bttn {
  margin-left:1rem;
}
```

bspeare:
Closed via https://github.com/uiowa/uids/pull/412.

PR #412, opened 2020-11-16
Banner Video Variant

GaryRidgway:
This PR is tracking all commits that are for adding the video  variant to the banner component similar as to how it is on the hero component.

## To Test

- [x] Test that the `Button Row` variant has buttons that individually work and do not link to the same thing as clicking the whole banner. Also ensure that the link in the text functions independently as well.
- [x] Test that the Banner can be unlinked while there are multiple buttons
- [x] Test that both the `Large, content centered ` variant and the `Button Row` variant have buttons that can be right clicked to open them in a new tab
- [x] Test that the Banner video works
- [x] Test that a banner video with a button row works
- [x] Test that the auto-play/auto-pause for banner video works
- [x] Test that the video is available and reasonably shown at all banner sizes
- [x] Test that there can be multiple videos and that you can play and pause them independently.


GaryRidgway:
This is specifically for issue #394

GaryRidgway:
The scope of this PR has now expanded to include #393.

bspeare:
Just checked https://uids.brand.uiowa.edu/branches/i394-banner-video/components/preview/banner--video.html on IE and I can confirm that it does not work.   Do we want this video loop to work with IE11?

bspeare:
I'm working on adding an IE style fix so that the banner will display as a normal image banner.   I think getting the video play/pause buttons to work in IE might be work for a featured sprint.

richardbporter:
> buttons that individually work and do not link to the same thing as clicking the whole banner.

Is that something we need to allow/support/implement? I don't think we can do that in Drupal. If one link is provided, the whole banner is linked. Otherwise, the banner is not linked.

GaryRidgway:
That should be accounted for in the JS, that is why i was accounting for two separate link fields. Perhaps we should have a 'banner link' field and then a multi-value 'button links' field? My initial implementation was trying to allow for as little change of old framework as possible.

richardbporter:
I did not get that impression from https://github.com/uiowa/uiowa/issues/2404. A separate field of one link and another of up to two links seems less than ideal for the editor. 

I thought linking the whole banner was an a11y workaround for two identical links?

pyrello:
On the Drupal side, this should consist of changing the limit on the current link field to allow for 3 links instead of 1. The idea is that the banner will function as fully clickable if, and only if, there is one button link. If there are multiple button links, then we cannot determine which link clicking the banner should direct you to. In that case, I think that the behavior of having the entire banner linked would be confusing to the end user.

joewhitsitt:
pulling down to see if I can help with the consolidation task discussed in slack

joewhitsitt:
uh, getting errors locally with npm run watch

joewhitsitt:
deleting node_modules is causing more errors. i might not be the knight in shining armor for this one.

joewhitsitt:
i still haven't been able to get a build to finish here. i might try to delete the repo locally and start fresh

joewhitsitt:
So, i'd like some eyes on my commits here. Alan said we could basically copy/paste banner.twig to banner-multiple if it is the correct approach and that he was going to get the js link working again based on the updated structure.

Is this the right approach?

pyrello:
I'm pulling this down to review the changes.

pyrello:
I have not done anything to resolve how we are going to move forward with the accessibility strategy that was adapted from cards, which was, in turn, adapted from this article: https://inclusive-components.design/cards/

I did, however, make two updates to simplify maintainability:
* Added a video template that can be included in the banner, rather than burdening the banner with that.
* Refactored the `banner--multiple.twig` template to use the existing single banner template.

GaryRidgway:
I've re-tested all of the new changes and they all seem to work great except for single buttons being able to be right clicked because of an accessibility quirk. upon discussion with both Sean and Ben This seems to be a minimally detrimental problem and not a real hinderance to the spirit of this pull request.

pyrello:
Great work on this @GaryRidgway and everyone else who pitched in.

bspeare:
Pulling this down to test.

bspeare:
Looks like we have one more issue to fix, a conflict between the `hero.js` and `banner.js`:

Screen Shot 2020-12-04 at 1 18 46 PM

pyrello:
@bspeare Can we just stop using hero.js?

bspeare:
@pyrello I think we would need to socialize that a bit more to stakeholders beforehand.  We can only load the `banner.js` if a video is uploaded in https://github.com/uiowa/uiowa/pull/2545?

bspeare:
The entire banner doesn't seem clickable when I have one link in place on https://github.com/uiowa/uiowa/pull/2609.  It works on https://localhost:3000/components/preview/banner--default, but not on the uiowa pr.  Do we need to update the banner template there?

pyrello:
@bspeare I changed the query selector for the title in banner.js to target a generic `.headline` class instead of the `bold-headline`. This was already present on the headline template here (which is why it worked), but hadn't been added to the template in UIDS Base yet.

bspeare:
@pyrello It looks like that did fix the issue.  I'll need to do some further testing with how the `.headline` class will have an impact with regards to how this is set: https://github.com/uiowa/uids/blob/main/src/components/headline/headline.scss#L58-L68.

bspeare (approved):
I've tested this locally and in the two pull requests in `uiowa`.  Nice work!

Issue #415, opened 2020-11-20
Add styling for <small> tag in typography component


bspeare:
# Problem

The `<small>` tag currently doesn't have a defined style in the typography component. 

# Proposed Solution

Add `<small>` tag styling with adjusted line-height in the typography component.

PR #430, opened 2020-12-07
Feature: small tag adjustments

bspeare:
Resolves https://github.com/uiowa/uids/issues/415. 

# How to test

Go to https://localhost:3000/components/detail/paragraph and verify that the small tag has a different `line-height` than the paragraph text.    Verify that the area of text with the class `small--inline` will display inline with the paragraph text with the same `line-height` as the paragraph.

### DIFF
--- a/src/components/typography/paragraph/paragraph.scss
+++ b/src/components/typography/paragraph/paragraph.scss
@@ -62,6 +62,8 @@ pre {
 small {
   vertical-align: middle;
   display: inline-block;
+  line-height: 1.3;
+
   &.small--inline {
     display: inline;
   }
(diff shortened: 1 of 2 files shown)


joewhitsitt (approved):

bspeare:
@joewhitsitt Do we need to add the `<small>` tag to the `filtered_html` settings?  Do you have any objections to that?

joewhitsitt:
> @joewhitsitt Thanks! Do we need to add the `<small>` tag to the `filtered_html` settings? Do you have any objections to that?

no objections @bspeare. surprised this hasn't happened sooner.

PR #431, opened 2020-12-07
add primary color for mark tag

joewhitsitt:
We currently allow the `<mark>` tag in our WYSIWYG. Reset.css makes it highlighter yellow. May I suggest an on-brand mark?

before:
Screen Shot 2020-12-07 at 2 24 45 PM

after:
Screen Shot 2020-12-07 at 2 24 34 PM


bspeare (approved):
Good idea!

Issue #391, opened 2020-10-26
Create tab styles and js within tabs component


bspeare:
# Problem

There is a tabs component started but not yet finished: https://github.com/uiowa/uids/tree/main/src/components/tabs. 

# Proposed Solution

Create new `CSS` and `js` that can be used to create a tabs component that can be marked as `WIP` and implemented in `uids_base`.

PR #434, opened 2020-12-08
Tabs Component

GaryRidgway:
This PR is to track the progress for #391.

This should be in a reasonable state for functionality, but we should just merge this so that we don't have hanging feature branches.

There are no styles yet, Functionality and accessibility were first priorities.


pyrello:
I think I have fixed the outstanding issues I created with refactoring. This looks good to me from the standpoint of being a solid foundation to work from later on. @GaryRidgway If you can double check when you are able that all the functionality that you would expect is there, I would appreciate it.

pyrello (approved):
Tabs function, but are without styling which is okay for now.

PR #435, opened 2020-12-08
Hotfix: Video updates

pyrello:
* Adds a check that the video container element and the video element exist before running the rest of the javascript.
* Converts Video object methods to be static since they don't need to reference `this`.

# To test
* Check that https://uids.brand.uiowa.edu/branches/hotfix_video_check/components/detail/banner--video still works.
* Check that https://uids.brand.uiowa.edu/branches/hotfix_video_check/components/detail/banner--multiple still works.


bspeare (approved):
Both links still work!

PR #442, opened 2020-12-18
Minor table updates

joewhitsitt:
- Handle table border attribute a bit better instead of stripping it out completely if provided.
- Add cost of tuition table example to run further table work against.

## Table border

- Viewing tables should look the same as they did before.
- Adding `border="0"` to any of the tables shouldn't have too much visual effect.
- Adding `border="1"` to any of the tables should have soft color borders unless it is overridden later by other styles.

## Cost of tuition example
Cost of tuition table is used by many sites. Any further table work should be able to handle this complex table in some fashion: https://uids.brand.uiowa.edu/branches/tuition-table-example/components/detail/tables.html.

In the wild examples:
https://financialaid.uiowa.edu/cost/undergraduate
https://admissions.uiowa.edu/finances/estimated-costs-attendance

https://uids.brand.uiowa.edu/branches/tuition-table-example/components/detail/tables.html

Screen Shot 2020-12-18 at 12 20 01 PM


pyrello:
It would be nice to have a right-aligned cell option, to allow those dollar figures to line up.

joewhitsitt:
> It would be nice to have a right-aligned cell option, to allow those dollar figures to line up.

Agreed, but I think it is out of scope for this PR. This was just a quick to establish a table to test/build against going forward.

`text-align-right` seems to work in Drupal but I am not sure where it is coming from.

bspeare:
Based on the tables conversation we had today and Todd's email, should we just make this the default table style in `uids_base` and create a class for a more minimal table style instead?

joewhitsitt:
> Based on the tables conversation we had today and Todd's email, should we just make this the default table style in `uids_base` and create a class for a more minimal table style instead?

Addressed this by looking for the border attribute to prevent borders if not present. Hopefully this satisfies the issue while keeping table borders from causing havoc.

bspeare (approved):
This looks good.  Do we want add a future issue for adding a utility class to align content?

Issue #443, opened 2020-12-21
As a developer, I can use a separate grid/flexbox utilities file to include to grid and flexbox mixins


bspeare:
# Problem

Currently, flexbox mixins exist in the `_utilites.scss` file and some new grid mixins are in `uids_base`: https://github.com/uiowa/uiowa/blob/fad9532999b2e4197773a8e4e74b135ba8bc6c49/docroot/themes/custom/uids_base/scss/components/grid.scss. 

# Proposed Solution

Create a `_grid.scss` utility file using:
- [ ] https://github.com/uiowa/uiowa/blob/fad9532999b2e4197773a8e4e74b135ba8bc6c49/docroot/themes/custom/uids_base/scss/components/grid.scss
- [ ] Move flexbox utilities into this new file: https://github.com/uiowa/uids/blob/main/src/assets/scss/_utilities.scss#L141-L185
  - [ ] Update components using the flexbox utilities.

PR #444, opened 2020-12-21
add grid mixin, move flex mixin

joewhitsitt:
Resolves https://github.com/uiowa/uids/issues/443

Locally, add some markup like:

```
<div class="parent-container">
  <div class="inner--grid grid--twocol--50-50">
    <div class="list-container">
      <div>item 1</div>
      <div>item 2</div>
      <div>item 3</div>
      <div>item 4</div>
      <div>item 5</div>
    </div>
  </div>
</div>
```
Somewhere in SCSS where the utilities partial is loaded, add:

`.parent-container {@include inner-grid(75);}`

A watch/build should result in grid items.

Also look around to make sure no existing flex work is broken as it has moved.


bspeare (approved):
This looks good to me.  @pyrello do you have any objections to merging this in?

bspeare:
One additional question on this.  Do we want to create a separate `grid.scss` component so that we can utilize these grid utility clases so that we can apply `inner--grid.grid--twocol--50-50` to a container wrapper or view and utilize these styles:
```
.inner--grid.grid--twocol--50-50 .list-container {
      display: grid;
      column-gap: 3rem;
      row-gap: 1.25rem;
      grid-template-columns: repeat(auto-fill, minmax(46.7%, 1fr)); 
}
```

joewhitsitt:
>We would probably need to add some additional mixins to accomplish this.

I might be confused about the use-case here. I built this with the parent column width in mind for block lists and assumes these classes are programmatically or template provided. You can add these classes to the view through the view configuration. If you just want a more generic grid css mixin, you could create something separate similar to the flex mixin that doesn't have the minmax adjustment and takes in custom gap values.

bspeare:
@joewhitsitt the use case here would be for views and displays created and placed outside of layout builder.  Like a view page.

bspeare:
Or, how do you think we should style a view page using this grid mixin?

Screen Shot 2020-12-22 at 9 05 18 AM

joewhitsitt:
>created and placed outside of layout builder

I wasn't thinking about this when I created this. Only about the different types of wrappers we encounter within layout builder with added lb_styles (aggregator-wrapper, view-content, etc.)

Short term can we add `inner-grid` and `grid--threecol--33-34-33` to the view(s) you are working on and create a separate issue that would have all the specifics of what you are thinking. I heard that Rick is also trying to wrangle. Is he able to control any markup or add these classes?

For me, I'd have troubles knowing where to stop with potential arguments for this mixin. (parent width, max-column-gap, max-row-gap, number of columns, max-width column specifics (e.g. 30-40-30)).

bspeare:
@joewhitsitt that sounds good.  I can create another issue with a proposed solution that can help with other types of displays like the student profiles view and Rick's APR work.

richardbporter:
Would it help to merge #2725 in sooner with any of this? Since the APR split and module is not enabled for anyone and won't be for awhile, the styling could wait as a follow up.

src/assets/scss/_breakpoints.scss
```diff
@@ -0,0 +1,55 @@
...
+  }
+
+  @else if $class==xlg {
+    @media (min-width: 106em) {
+      @content;
+    }
+  }
+
+  @else {
+    @warn "Breakpoint mixin supports: xs, sm, md, lg";
```
pyrello, line 53: Should this include the other options that are supported?

bspeare: Good point!

pyrello (approved):
Reviewed, not tested.

PR #454, opened 2021-01-12
Template styles

GaryRidgway:
this adds styles for the new vue template project.

The core structure of these things should be changed eventually to mirror the new classes, but for now this will work.


bspeare (approved):
Reviewed, not tested.

PR #458, opened 2021-01-12
Feature: Fix for stat hover regression

bspeare:
This pr fixes a stat regression that loses it centering if the stat contains `stat__content`. 

# How to test

Compare https://uids.brand.uiowa.edu/components/preview/stat--row.html against https://uids.brand.uiowa.edu/branches/feature_stat_regression_updates/components/preview/stat--row.html and notice how the hover in this feature branch does not pop all the way to the top of the grid container.


pyrello (approved):
Looks good.

PR #479, opened 2021-02-03
horizontal nav fix

GaryRidgway:
This Pr Fixes a style problem with Horizontal menus at lower breakpoints.

# To Test
1. Pull down this branch 
2. Change the `hidden: true` value on the second line of `src/components/menus/main-menu/main-menu.config.yml` to `hidden: false` (Do not commit this)
3. Spin up your local site with `npm run watch`
2. Under Menus > Main Menu check that all horizontal styles at less than 768px have no down arrow, and that the tab/arrow styles only appear at greater breakpoints.

What you do not want to see:

Screen Shot 2021-02-03 at 2 06 06 PM

What you do want to see: 

Screen Shot 2021-02-03 at 2 07 13 PM

### DIFF
--- a/src/components/menus/main-menu/main-menu.scss
+++ b/src/components/menus/main-menu/main-menu.scss
@@ -127,7 +127,7 @@
     li {
       @include breakpoint(sm) {
         font-size: calc(100% + .1vw);
-        @include center(vertical);
+        display: flex;
       }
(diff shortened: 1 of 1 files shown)


bspeare (approved):
Looks good, thanks!

Issue #468, opened 2021-01-20
As a site owner, I can add additional style and formatting options to my tables


bspeare:
As a site owner, I can add additional style options to my tables so that I can present my tabular data beyond what the basic table styles provide. 

# Proposed Solution

- [ ] Audit current table styles and compare against the list of tables below
- [ ] Provide a list of missing options to implement within the table component
- [ ] Implement the options within the component 
- [ ] **Optional:** any attributes that would need to be added to https://github.com/uiowa/uiowa/issues/2826
- [ ] https://github.com/uiowa/uids/issues/487

https://admissions.uiowa.edu/instructions/undergratuate-domestic

<img width="1232" alt="Screen Shot 2021-01-20 at 1 40 30 PM" src="https://user-images.githubusercontent.com/1036433/105226123-2f468380-5b25-11eb-9bc2-09138b71bfd0.png">

https://admissions.uiowa.edu/nursing-technical-standards

<img width="1029" alt="Screen Shot 2021-01-20 at 1 41 52 PM" src="https://user-images.githubusercontent.com/1036433/105226181-484f3480-5b25-11eb-8031-9bc94f52a411.png">

https://admissions.uiowa.edu/you-ui-schedule-test

<img width="1029" alt="Screen Shot 2021-01-20 at 1 41 52 PM" src="https://user-images.githubusercontent.com/1036433/105226235-5b620480-5b25-11eb-89b0-0e9a5cd8e9a1.png">

https://admissions.uiowa.edu/academics-undergraduate-programs-undergraduate/actsat-equivalency-table

<img width="1357" alt="Screen Shot 2021-01-20 at 1 42 27 PM" src="https://user-images.githubusercontent.com/1036433/105226293-6cab1100-5b25-11eb-935e-32dadcb049be.png">

https://admissions.uiowa.edu/academics-undergraduate-programs-undergraduate/minimum-high-school-course-requirements
https://admissions.uiowa.edu/future-students/high-school-course-requirements-international-applicants

<img width="837" alt="Screen Shot 2021-01-20 at 1 43 15 PM" src="https://user-images.githubusercontent.com/1036433/105226327-792f6980-5b25-11eb-8d8f-6a3344255ad9.png">

https://admissions.uiowa.edu/undergraduate-programs/international-undergraduate-application-deadlines

[...]

PR #490, opened 2021-03-09
table styles for admissions

GaryRidgway:
This pr is for  #468 

# How to test:
1. pull down this PR and spin up your instance of the UIDS
2. test that the following exist:
    1. Adding a caption to the table will show a grey caption above the table itself.
    2. Adding the `.gray-borders` class will put gray borders on a table.
    3. Adding the `hover-highlight` class allows rows to be highlighted on hover.
    4.  Adding the `table--width-default` class allows a table to shrink to only as wide as necessitated.


joewhitsitt:
pulling down to test

joewhitsitt (approved):
Works as outlined. It would be nice to polish the table border attribute in the future to act similar to grey-borders (see what i did there?) since that is allowed as an a11y feature.

Allowing some of this on table--static would also be a nice to have in the future.

PR #491, opened 2021-03-11
fixing table striped border issue

GaryRidgway:
hotfix to fix sizing issue. Borders were being added where they were not needed, and making the width calculations mess up.


pyrello (approved):
Reviewed, not tested.

PR #495, opened 2021-03-17
Feature table style fixes

GaryRidgway:
This PR is to adjust some of the table styles for borders and to BEM the class names


bspeare (approved):
Thanks for the refactor!

PR #498, opened 2021-03-19
Remote hosting and file cleanup

GaryRidgway:
This PR fixes the inability for users to click the pause button on banner videos and also starts to try and remotely host images instead of host them in our repo, freeing up space.


src/components/banner/banner.scss
```diff
@@ -82,6 +82,7 @@
     margin: 0 auto;
     width: 100%;
     z-index: 1;
+    pointer-events: none;
```
bspeare, line 85: This removes the hover on the buttons attached to the banner.   I think just a `z-index` adjustment will fix this issue instead:
```
.media--video .video-controls {
    z-index: 2;
}
```

bspeare (approved):
Great cleanup!  Approved as long as the banner buttons will still show a hover effect.

PR #504, opened 2021-03-25
fixing table width

GaryRidgway:
This PR should reset all tables to a default of 100% width, with `table--width-default` opting out of the 100% width.

# To test
1. Pull down this branch
2. run `npm run watch` to spin up the server
3. check that if a table is given `table--static` it will be 100% width until it is given `table--width-default`


joewhitsitt (approved):
Tested. Width 100% tables by default is how we have operated for a long time and what I approved in https://github.com/uiowa/uids/pull/490. I know it is technical debt and not default HTML spec, but we can't pull the rug out suddenly by switching this behavior. Thanks for the quick change.

PR #518, opened 2021-04-12
Feature: Default table styles

### LINKED ISSUE #509: Style Tables with default styles.
Part of the [507 Responsive Tables](https://github.com/uiowa/uids/issues/507) epic.

Style tables with default styling closer to what we would like to see for tables. this does not include a side scrolling option. Responsive tables will not be removed during this time.

# Proposed solution

Go through tables and make the base table styles striped and with gold <th> tags.

### LINKED ISSUE #510: Set an overflow solution for default tables
Part of the [507 Responsive Tables](https://github.com/uiowa/uids/issues/507) epic.

Default tables need to have a simple horizontal overflow option that allows them to be used in thin widths.

# Proposed solution

1. Remove responsive functionality from triggering on any table.
2. Style Tables with a simple overflow horizontally

pyrello:
Resolves #509
Resolves #510 

# How to test

https://uids.brand.uiowa.edu/branches/feature_default_tables/components/detail/tables.html

Check different table examples. The previous responsive behavior has been removed in favor of a horizontal scroll that is in effect below the `sm` breakpoint. Sticky headers (top or left) will no longer work. Compare the tables with the current tables to see if anything appears broken in appearance.

There is a hidden example of using a wrapping div for adding horizontal scrolling to larger tables beyond the `sm` breakpoint. To view it, set the responsive variant to not be hidden locally in `src/components/tables/tables.config.yml` and view the resulting page that is added to the documentation.

### DIFF
--- a/.github/workflows/gh-pages.yml
+++ b/.github/workflows/gh-pages.yml
@@ -65,7 +65,7 @@ jobs:
 
       # Debugging messages.
       - run: echo "${{ github.ref }}"
-      - run: echo "${{ toJson(github) }}"
+#      - run: echo "${{ toJson(github) }}"
 #      - run: echo "${{ env.MAIN_BRANCH }}"
 #      - run: echo "Main check - ${{ github.ref == format('refs/heads/{0}', env.MAIN_BRANCH) }}"
 #      - run: echo "Branch check - ${{ contains(github.ref, 'refs/heads/') && github.ref != format('refs/heads/{0}', env.MAIN_BRANCH) }}"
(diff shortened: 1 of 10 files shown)


bspeare (approved):
Looks good!

GaryRidgway:
Should it be `table--responsive` instead of `table-responsive`? that feels a bit closer to our current BEM strategy.

pyrello:
> Should it be table--responsive instead of table-responsive? that feels a bit closer to our current BEM strategy.

Good question. Since the table wrapping div is not strictly part of the table and that class is really more of a utility class than anything else, I was thinking that not using BEM naming made more sense in this case.

GaryRidgway:
that makes sense, do we want to use that as a prefix for any responsive only classes in the future since they will be sub classes of `table-responsive`?

pyrello:
I think that's a good idea, Alan!

Issue #531, opened 2021-04-19
Create `card--nolink` functionality in card.js


bspeare:
As a part of https://github.com/uiowa/uiowa/issues/3262, we need an opt-out class for cards that don't need the entire card clickable/linked. 

# Proposed solution

- Create `card--nolink` or similar modifier class to target in `card.js`. 
- Change https://github.com/uiowa/uids/blob/main/src/components/card/card.js#L1 to be something like `const cards = document.querySelectorAll('.card:not(.card--nolink)')` ?

GaryRidgway:
This makes sense to me, and i think it's a pretty consistent implementation with other things that we have done before.

PR #532, opened 2021-04-19
Feature: Added data attribute for removing entire card link function in card.js

bspeare:
Resolves https://github.com/uiowa/uids/issues/531. 

# How to test

- Go to https://uids.brand.uiowa.edu/branches/feature_card_no_link/components/preview/card--row.html and verify that first card has data attribute and that only the card title is clickable.


pyrello (approved):
Tested. Works as expected.

Issue #538, opened 2021-05-03
Fix `tables--width-default` class behavior.


pyrello:

bspeare:
Looks like `tables--width-default` might just need `width: auto;` set in order to allow tables to be narrow.

PR #539, opened 2021-05-04
Added a fix to caption with to work with table--width-default class a…

bspeare:
Resolves: https://github.com/uiowa/uids/issues/538

# How to test

- If the pages generate, check out https://uids.brand.uiowa.edu/branches/feature_table_hotfix/components/preview/tables.html or https://localhost:3000/components/preview/tables and look for the table with the heading "Table with default width" has a caption that does not extend full width of the page.


pyrello (approved):
Reviewed. Thanks!

PR #552, opened 2021-05-14
Adjust card and banner click handling JS

pyrello:
Related to #553, but does not resolve


pyrello:
I have adjusted this to include banners by moving the logic into a shared assets JS file and adjusting the implementation to use component-agnostic classes. Also just letting `<a>` and `<button>` clicks do what they would have originally have done and only applying the click behavior otherwise.

richardbporter:
Adjusted PR description to not resolve the original issue.

joewhitsitt (approved):
Approving knowing that this is a step forward but does not address https://github.com/uiowa/uids/issues/553

pyrello:
I've closed #553 as "won't fix" based on my previous investigation.

PR #563, opened 2021-05-27
Intro Text Changes

bewoldt:
Proposed changes to Intro Text feature.

- Smaller font size. Feels odd to have the intro text larger than the H2.
- Increased line-height for spacing
- Removes additional margin-bottom and defaults to regular paragraph margin-bottom.

Before
Screen Shot 2021-05-27 at 10 24 10 AM

After
Screen Shot 2021-05-27 at 10 50 28 AM

### DIFF
--- a/src/components/typography/paragraph/paragraph.scss
+++ b/src/components/typography/paragraph/paragraph.scss
@@ -164,15 +164,15 @@ pre {
   font-weight: $font-weight-medium-bold;
   line-height: 1.3;
   font-size: 2.0rem;
-  font-size: containerClamp(1.6rem, 2.0rem);
+  font-size: containerClamp(1.6rem, 2.2rem);
   margin-top: 0;
 }
 
 .uids-component--light-intro,
 .element--light-intro {
   font-weight: $font-weight-light;
-  line-height: 1.3;
+  line-height: 1.4;
   font-size: 2.0rem;
-  font-size: containerClamp(1.6rem, 2.0rem);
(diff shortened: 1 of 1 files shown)


bspeare:
@bewoldt This seems like a good change.  Would you be willing to move up to `2.1rem`?

bewoldt:
> @bewoldt This seems like a good change. Would you be willing to move up to `2.1rem`?

@bspeare 2.1rem is fine.

bspeare:
@bewoldt I adjusted the line-height a bit as well and adjusted the font size for the bold option because Zilla has a smaller height than roboto.  Can you review the changes?

bspeare:
Also, there is work happening in https://github.com/uiowa/uids/pull/561 that will increase the `<h2>` size for Zilla so that it has a similar height to the Roboto `<h2>`.

bewoldt:
I still think it's slightly larger than it needs to be but definitely better than it was before.

bspeare:
OK, let's wait until https://github.com/uiowa/uids/pull/561 merges and then adjust as necessary.

bewoldt:
I just discover you have a .lead class name too. Is the Intro Text meant to replace this? Are you keeping .lead? https://pharmacy.uiowa.edu/pharmacy_practice

bspeare:
We intend to replace `.lead` with this set of styles.

bspeare (approved):
Let's go with this and revisit in a future pull request if we still aren't happy with it.

Issue #556, opened 2021-05-24
Add Storybook to UIDS


pyrello:
# Motivation

Storybook is a development tool that allows for easily building and testing components. It has a "knobs" feature that allows you to change settings on a component (such as background color or text orientation) without having to build separate example components for every combination of variations. It would be a powerful tool to ensure better outcomes when testing new features and changes to existing features.

bspeare:
Adding this here in case we need it https://storybook.js.org/blog/storybook-for-vite/.

pyrello:
This was done in #673.

PR #578, opened 2021-06-11
Beginning of Storybook integration

pyrello:
Starting to address #556


pyrello:
This already happened in #673.

PR #584, opened 2021-06-29
Feature: Sticky table headers

bspeare:
https://uids.brand.uiowa.edu/branches/feature_sticky_table_headers/components/preview/tables--responsive.html

This method requires adding a modifier class to the `<div>` around the table.   This `<div>` is being added by a module in Drupal, so we would need to take that into account when we implement it into the service. 

# Remaining tasks
- [x] https://github.com/uiowa/uiowa/issues/3576#issuecomment-870850320
- [x] Apply `position: sticky` to the `<thead>` or `<tr>`?


joewhitsitt:
Looks pretty good to me. Tested in safari, chrome, firefox.

Issue #424, opened 2020-12-02
Slider hover is rubberbanding and causing it to shrink less than its intended width.


GaryRidgway:
## PROBLEM
As stated in the title, this is what is occurring: 
https://i.gyazo.com/5c15022585159eb20ad1bb49de548922.gif
I believe this is happening because different images are being hovered too fast, making all of the transitions happen at the same time, shrinking all the containers at once.

## PROPOSED SOLUTION
Find a way to interrupt the animations on hover

OR 

Allow for overflow on the right side of images and allow them to expand in to the container beyond their width but not outside the parent container so that even if they are too small there is still some image overflowing on the right side to fill the space.

PR #616, opened 2021-12-20
add transition offset to block rubberbanding lag

GaryRidgway:
this should fix https://github.com/uiowa/uids/issues/424


bspeare (approved):
Great fix! I see no more rubberbanding!!

Issue #629, opened 2022-02-03
Finish tabs component


pyrello:
# Motivation

It appears we started working on a tabbed component at some point, but never finished it: https://uids.brand.uiowa.edu/latest/components/detail/tabs.html. The JS seems to function correctly, but the styling is not finished.

# Proposed solution

Use this screenshot as a starting point. 

<img width="2068" alt="Screen Shot 2022-02-09 at 1 43 13 PM" src="https://user-images.githubusercontent.com/1036433/153278149-f44c91b1-7fca-468a-8dcf-dda50af38e83.png">

makurj:
When Alan picks this up :), I would like to pair on this.

bspeare:
Do we want to be able to open tab items based on url anchor like https://github.com/uiowa/uids/issues/396?

PR #637, opened 2022-02-11
Tabs component styling changes

makurj:
Tracks #629 

# How to test

- Test with `bg--black`, `bg--gold`, and `bg--gray` background classes on the `<body>` and verify that tab hover/active states remain with accessible color contrast. 
- Test that you can navigate with keyboard only and verify that focus styles provide sufficient indication of where you are within the tabs.


makurj:
Huge thanks to @GaryRidgway & @bspeare for all their help!

bspeare:
I think the styling part of this is done but we might need to figure out whether https://github.com/uiowa/uids/issues/629#issuecomment-1042047396 is working for this and what needs to be done to get it working.

makurj:
The work above will only work in the component preview.

bspeare:
I see that the url linking to the tabs works by grabbing the generated url string like https://uids.brand.uiowa.edu/branches/feature-tabs/components/preview/tabs--default?tabs--0=tabs--0__tab--1, I wonder if the expected behavior is to link to a section with a link constructed like https://uids.brand.uiowa.edu/branches/feature-tabs/components/preview/tabs--default#tabs--0__tab--2?

GaryRidgway:
so currently the way we have it is able to track multiple tabs groups on the page. does this new implementation assume that there is only one tab group that we want to track? i don't know how we would do multiple with the id URLs

joewhitsitt:
could it just be simplified then? `?tabs--0=t0tab--1` or something?

GaryRidgway:
> could it just be simplified then? `?tabs--0=t0tab--1` or something?

It's ID based, it will be whatever drupal sets as the IDs

GaryRidgway:
I guess we could make it not ID based? i hadnt thought of doing that but it is possible to simplify. maybe `?t0=t1`?

GaryRidgway:
The benefit of the IDs in the url however is that it required no iteration through the Dom to get what tabgroup we are looking for, because we can just put the id in to a query and get it immediately, instead of loading every tabgroup and iterating through until we find the correct one.

bspeare:
Maybe the first slice is to get the `#` working on a single tab and we can revisit for multiple tabs?  I think we would want to use the standard hash for linking to these tabs since that's a common pattern you will find with other design system tabs.

PR #642, opened 2022-02-22
Hash tabs

GaryRidgway:
Changing from query strings to has queries

tracks https://github.com/uiowa/uids/pull/637

## To test:

1. Go to the component preview of a tabs component. You must test this way because the normal display is an iframe, and cannot get arguments from the URL properly.
2. Click around and observe the URL changing when different tabs are clicked.
3. Use the arrow keys to navigate and observe the correct tab being added in the URL when the tab changes in the tabgroup.
4. Copy any paste a URL with a hash parameter in it in to a new tab, and observe that the tab specified in the hash is the one that is automatically focused and opened.
5. Try and break it and post any concerns


richardbporter (commented):
Works for me. Is this PR replacing #637? Should that be closed then?

GaryRidgway:
yes, this replaces that. I will close that, but we wanted the branch out there in case we wanted the code to transfer over for functional reasons ever

Thanks Rick!

Issue #396, opened 2020-10-30
Open accordion item based on URL anchor


joewhitsitt:
## Issue

Accordions have a unique id that can be targeted (e.g. https://sitenow.uiowa.edu/faqs#heading351), but doesn't currently open the item.

## Proposed Solution

Make it happen!

bspeare:
Related issue: https://github.com/uiowa/uids/issues/112

GaryRidgway:
This should probably include some of the enhancements that we made for the hashTabs such as:

- [ ] displaying all data when no JS or css is present
- [ ] changing the open accordion based on a change in the URL
- [ ] Being able to set any ID you want and have it still work

PR #643, opened 2022-02-25
Hash tabs JS fallback

### LINKED ISSUE #657: Tabs enhancements: Readable hashes
Change the hashes to not need a specific format to be usable.

As a tabs user, i can set any id on my tabs and they will work with the code that has already been implemented.

tracked in https://github.com/uiowa/uids/pull/643

GaryRidgway:
Resolves #657 

Is to make Hash tabs work without JS and CSS.

## To test:
1.  Turn off JS and CSS
2. check if you can still see all the content of the tabs
3. make sure that you are still moved down the page to the proper tab (You can test this by adding a lot of paragraphs to the template.)

### DIFF
--- a/src/components/tabs/tabs--focus-test.twig
+++ b/src/components/tabs/tabs--focus-test.twig
@@ -1,0 +1,35 @@
+<div id="lipsum">
+  <p>
+    Lorem ipsum dolor sit amet, consectetur adipiscing elit. Morbi malesuada urna vitae convallis semper. Pellentesque habitant morbi tristique senectus et netus et malesuada fames ac turpis egestas. Ut lobortis hendrerit molestie. Aenean porttitor suscipit dapibus. Sed aliquet auctor scelerisque. Vestibulum vehicula urna eget nisl rhoncus, non cursus mauris tristique. Pellentesque at accumsan metus. Integer vitae nulla vitae metus semper volutpat id nec quam. Fusce iaculis aliquet sapien quis commodo. Nullam risus tortor, commodo vitae molestie ac, consectetur non eros. Etiam pharetra quis lectus vel semper. Nam nec lacus dolor.
+  </p>
+  <p>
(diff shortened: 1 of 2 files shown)


pyrello:
I missed the first round of this, so apologies that I am coming to this issue late. I'm trying to follow everything that has happened thus far which has led to the need for this PR.

It appears that [a comment](https://github.com/uiowa/uids/issues/629#issuecomment-1042047396) was added to the issue to finish the tabs component by @bspeare:
> Do we want to be able to open tab items based on url anchor like https://github.com/uiowa/uids/issues/396?

That comment points to #396 (which is a bit vague in its details), but is basically an issue to allow for anchor links to trigger an accordion to open.

I guess my expectation of how this would work is that when an anchor link is passed, as in example.com/test#gohere, that the page would jump to the correct place on the page and if that anchor also happened to be a tab, it would trigger the tab to open.

It seems like the code is doing a lot more than that. My concern here is that we are maybe putting the cart before the horse by developing a more robust solution before we already have that basic use case met, working, and vetted by usage on a number of sites.

bspeare:
I tried recreating the issue noted above and I could not reproduce the issue.  What version of Brave are you using @pyrello?  Did this occur in other browsers?

bspeare:
I was able to recreate the bug. 

I had https://localhost:3000/components/preview/tabs--default#oatmeal open in Firefox or Brave, and then I changed the hash to `#sugar` in the url bar and hit enter.   Only after a second enter did the tab change.

GaryRidgway:
So I discovered that changing the hash parameters of a URL doesn't trigger a submit to the browser the first time, so I added in a listener that checks for when the URL is changed and then focuses the right element.

bspeare:
@GaryRidgway that fixed the issue for me.  It now instantly switches to the tab without having to hit enter twice.

pyrello:
Last thing I would like to see here is a way of testing that the page actually jumps to the tab rather than just switching it on. I think the best way to do this is to add a (hidden) example with lorem ipsum text before and after to allow the viewport to move to the tab that is being triggered. Otherwise this is looking like a very good enhancement.

pyrello:
Converting to draft until the above request is handled.

GaryRidgway:
Added, you should now be able to see it focus the tabs and scroll to them by navigating to https://localhost:3000/components/preview/tabs--focus-test#snickerdoodle

pyrello (approved):
Nice adjustments here. Thanks!

PR #658, opened 2022-03-11
refactor accordions to accept hash parameters

GaryRidgway:
This should start to be able to allow accordions to have hash parameter functionality.

## To test
- On https://localhost:3000/components/preview/accordion--default , test that clicking on different accordion items changes the hash parameter in the URL, and that closing an item clears the URL hash parameter.
- On the same page, check that changing the URL to another viable hash parameter also opens that accordion.
- Open a new tab and past in a URL of the page you just visited with a hash parameter and verify that the correct accordion item opens.
- On the same page, disable JS and CSS and verify that you can see all the accordion data, even if it is ugly.
- On https://localhost:3000/components/preview/accordion--multi , verify that clicking on different accordion items changes the hash parameter in the URL, and that closing an item clears the URL hash parameter. 
- On On https://localhost:3000/components/preview/accordion--multiple , verify that the new `id`'s do not collide or create any problems. There should be support for custom `id`s.
- On https://localhost:3000/components/preview/accordion--focus-test#snickerdoodle , check to see that you are brought to the open accordion item.
- On any page, try to break things and report any breaking here! (THANK YOU)

### DIFF
--- a/src/components/accordion/accordion--focus-test.twig
+++ b/src/components/accordion/accordion--focus-test.twig
@@ -1,0 +1,36 @@
+<div id="lipsum">
+  <p>
+    Lorem ipsum dolor sit amet, consectetur adipiscing elit. Morbi malesuada urna vitae convallis semper. Pellentesque habitant morbi tristique senectus et netus et malesuada fames ac turpis egestas. Ut lobortis hendrerit molestie. Aenean porttitor suscipit dapibus. Sed aliquet auctor scelerisque. Vestibulum vehicula urna eget nisl rhoncus, non cursus mauris tristique. Pellentesque at accumsan metus. Integer vitae nulla vitae metus semper volutpat id nec quam. Fusce iaculis aliquet sapien quis commodo. Nullam risus tortor, commodo vitae molestie ac, consectetur non eros. Etiam pharetra quis lectus vel semper. Nam nec lacus dolor.
+  </p>
+  <p>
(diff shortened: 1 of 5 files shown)


joewhitsitt:
going to try testing this

joewhitsitt:
>On the same page, disable JS and CSS and verify that you can see all the accordion data, even if it is ugly.

First, I forgot how to disable CSS without that chrome extension that was later flagged as being malware. Second, if just Javascript is disabled do we want to try and expose all accordion items? Scope creep for another day or do we just think JS is a prerequisite these days?

I am seeing `single-select-` for both "types" and it isn't getting the item_number appended to it, so all of the aria-controls are `single-select-`. I think something needs to be adjust here.

GaryRidgway:
> if just Javascript is disabled do we want to try and expose all accordion items?
This is currently the case, the items should be exposed if there is no JS.

> I am seeing `single-select-` for both "types" and it isn't getting the item_number appended to it, so all of the aria-controls are `single-select-`. I think something needs to be adjust here.
I will take a look!

> Yeah, I don't think all of the `item_number` were replaced with `accordion_item_number`. Also I am seeing mixed use of `variant` and `accordion_type`
I will also take a look here!

Thanks for taking a gander Joe!

GaryRidgway:
@joewhitsitt the issues you made visible should be fixed, would you mind taking another look?

joewhitsitt (approved):
Looks good to this casual UIDS reviewer. Thanks for sorting out the id/aria quickly.

PR #659, opened 2022-03-16
changing class, better checking for proper structure. Check the description for more details.

GaryRidgway:
Check the title for more details.


cory-skeers (approved):
Tested these changes locally, and they work like a charm! Think this will be good collision-proofing for possible future changes in core/contrib modules, too.

PR #698, opened 2022-05-27
Lets have a card

### LINKED ISSUE #696: 4.x - Create a card component
# Motivation

Create a card component to handle what we were using the card for in 3.x.

# Proposed solution

How are we going to reference different card sub elements (media, title, details, etc.) style-choices?
Elements
- [x] Media - The media options are going to be handled via `.media--` classes.
   - [x] ~Size~ 
   - [x] ~Shape/ Aspect Ratio~
   - [x] ~Outline~
- [x] Title
- [x] #708 - We are going to handle this in a follow-up issue.
- [x] Body
- [x] Footer/Actions

Options
- [x] Outline
- [x] Stacked/Alignment
    - [x] change the orientation of the default card, without the stacked class, to be horizontal.
- [x] Content alignment
- [x] Background
- [x] #709 - We are going to handle this in a follow-up issue.

pyrello:
Resolves #696 

# To test
- Visit https://uids.brand.uiowa.edu/branches/lets_have_a_card/
- Review card functionality against items in #696.
- Click through each card example and then toggle the different settings for each one to see if you can find any options that don't look right.
- I removed the container around the card because it was getting in the way of testing the media alignment options. I recommend using the viewport button to set it to a smaller width for testing the default card without a media alignment option selected.


joewhitsitt:
👀

joewhitsitt:
reviewing

joewhitsitt (approved):
In terms of the card (minus media), I tried several combinations and each seem to work as I expected however, a couple stood out to me and I am not sure if they are important or not:

When I click [`basic-card--linked-card-without-title`](https://uids.brand.uiowa.edu/branches/lets_have_a_card/?path=/story/basic-card--linked-card-without-title&args=link_text:) (removed link text) - I assume that is for SiteNow or other implementation to validate and not storybook.

When I clicked [`basic-card--borderless-card`](https://uids.brand.uiowa.edu/branches/lets_have_a_card/?path=/story/basic-card--borderless-card&args=media_padded:true) (mobile sized), I didn't notice a change in media padding. This was discussed in the huddle I was in earlier, but I am foggy about what was decided here.

Generally curious about how components will stack in Storybook and if we will be able to modify media or button settings while viewing the card component.

There is a lot of other stuff in this PR that I assume is safe to ignore for the time being since it wasn't the purview of this PR.

The fact that this can be reviewed in browser at this time and that certain card properties are handled, I think this is a good step forward!

PR #715, opened 2022-06-24
Update .nvmrc

joewhitsitt:


joewhitsitt:
We are now running node 18 here https://github.com/uiowa/uids/blob/4.x/.nvmrc

Issue #710, opened 2022-06-16
`bttn--tertiary` loses background color when used with `bttn--outline`


bspeare:
# Problem 

`bttn--tertiary` loses background color when paired with `bttn--outline`. 

# Proposed solution

Utilize `bttn--transparent` as a modifier to remove background colors of any button, like in the instance of `bttn--tertiary` using `bttn--outline`.

pyrello:
Per offline conversation, we discussed having the transparent option be included as a background color for the button. This is similar to what we are doing in #696. We also discussed making outlined buttons the default going forward, also similar to the card work happening in the 4.x branch.

@bspeare Do you think it might make more sense to add this to #706 and focus on getting uiowa/uiowa#5379 figured out so we can start using the new version of the button in SiteNow?

bspeare:
@pyrello that's a good idea to work in https://github.com/uiowa/uids/issues/706, is there way to test the button styles with nested background colors in the Storybook component?  Working within the `3.x` kitchen sink page is really useful for making changes to the button styles.

bspeare:
Closed via https://github.com/uiowa/uids/issues/706.

Issue #714, opened 2022-06-22
Add global `border` utility class


bspeare:
Add global `border` utility class.  This also might need some scoping to work with other components like accordions. 
```
.border {
  border: 1px solid;
  border-color: rgba(0,0,0,.425);
}
```
# Proposed solution
- Add new class listed above. 
- Remove class from https://github.com/uiowa/uiowa/blob/main/docroot/themes/custom/uids_base/scss/global.scss and https://github.com/uiowa/uiowa/blob/main/docroot/themes/custom/uids_base/scss/components/accordion.scss when this is complete.

Issue #100, opened 2020-04-05
Add /dist folder to .gitignore


pyrello:
Similar to USWDS: https://github.com/uswds/uswds/blob/develop/.gitignore#L14

pyrello:
This is maybe related to #73

richardbporter:
I'm not sure we can ignore dist/ since the uids_base theme uses that. I can look into changing the build and release process to make it a bit easier.

richardbporter:
This feels like it should be a part of #79 along with the folder restructure. However, there may be a fix in the short term. Thoughts?

pyrello:
@richardbporter The idea with adding `/dist` to .gitignore is that it is automatically generated by running `npm run build`, so presumably, it would get built as part of our automated build process. Not that it would be removed from the project as a resource.

richardbporter:
Our current automated build process pushes to the gh-pages page, basically for documentation purposes. Yarn/NPM get the package information from the tags created off master. If we ignore dist/ and build/ on master, they won't be included in the package.

I can take a closer look at this at some point but right now https://github.com/uiowa/uiowa/issues/1281 is my priority.

pyrello:
Does this line not run a build process that creates the `/dist` folder? https://github.com/uiowa/uids/blob/master/.travis.yml#L15

pyrello:
I have an idea for how to handle this. It would involve a new npm command to create a new release. Currently, we run an npm command: `npm version {major|minor|patch} -m "Message"`. This new command could run the build process and include the build directory, bump the version and tag a new release. This would be the "official" way of creating a new release that insured that the "/dist" folder was always included.

pyrello:
Here is an example of a release gulp task in USWDS: https://github.com/uswds/uswds/blob/develop/config/gulp/release.js

It looks like they are using CircleCI instead of Travis, but the format seems pretty similar. That command gets called as part of their deployment process: https://github.com/uswds/uswds/blob/develop/.circleci/config.yml#L61

pyrello:
It looks like their git workflow is a little different than ours: https://github.com/uswds/uswds/wiki/Release-process#git-workflow

In their workflow, the `master` branch is only updated at the point when a new release is created. They don't seem to use a "stable but not released" branch (our `master`). It looks like `develop` is more like our master branch.

bspeare:
@pyrello is this still an issue?

pyrello:
[...]

PR #716, opened 2022-06-27
4.x button updates

### LINKED ISSUE #706: 4.x - Button component updates
# Motivation
The [Brand Icon Browser](https://github.com/uiowa/brand-icon-browser) needs an exportable button component that has the following features, most of which are available in [the current button component](https://uids.brand.uiowa.edu/components/detail/button--row.html):

- Background color options
- The ability to have an appropriately styled Font Awesome Icon
- A possible "block" size option that takes the full width of the container, like the following example:
Screen Shot 2022-06-09 at 1 54 27 PM

# Proposed solution
- [x] Incorporate a transparent background option. The transparent option needs to have an outline, otherwise the text and icon will just be floating in space.  
  - Use https://github.com/uiowa/uids/pull/711 as a starting point for base CSS. 
    -  https://github.com/uiowa/uids/blob/acc28e4f0aae5ad5b92adb3d4871d941baf808e4/src/assets/scss/_utilities.scss
    - https://github.com/uiowa/uids/blob/acc28e4f0aae5ad5b92adb3d4871d941baf808e4/src/components/button/button.scss
- [x] Add an `icon` prop that allows for the display of an arbitrary FA icon. This will likely replace the `arrow` prop.
- [x] Add a `block` prop to set display of button to be full-width.
- [x] Abstract the outline prop and related logic from the Card component into a mixin which can be shared with the button.
- [x] Change borderless class to not be element specific

```
Props:

Color:
- Primary (gold)
- Secondary (black)
- Tertiary (white)
- Transparent (with border/outline added by default)
- Link (border removed no matter what)

Borderless:
- False/True
```

GaryRidgway:
Resolves #706
Resolves https://github.com/uiowa/uids/issues/710

## To test
- check to see that the transparent option is present, and that the transparent borderless option has link underlining.
- test that going to font awesome and copying an icon ([An example icon is here, copy everything in the `<i>` tag](https://fontawesome.com/v5/icons/dog?s=solid))
- Test that the full width option is available and working properly
- check that the borderless option works for both the card and button
- Test background color combinations, if possible https://github.com/uiowa/uids/pull/716#issuecomment-1170507596


pyrello:
@GaryRidgway It seems like the last thing you need to do here is write up testing instructions.

quamsta:
I'll likely do a test run on adding this to the Brand Icon Browser today, not sure if that counts as a full review, but I'll put anything of note in here.

quamsta:
@pyrello , @GaryRidgway  does the building/compiling of the `uids` module's  `dist/` folder happen during a GitHub action? I attempted to import the components in the Brand Icon Browser, but I quickly realized that the statement I was using:

`import { UidsBrandBar, UidsButton } from "uids";`

...was still importing the old version of the `UidsButton` component. Does the `dist/` folder need to be built in the `4.x-button-updates` branch or will it happen automatically when it's merged into `4.x`?

pyrello:
@quamsta This process is not automated yet :|

Here is the relevant issue: https://github.com/uiowa/uids/issues/100. Running `yarn build` should take care of it. Sorry we forgot to do this the other day.

quamsta:
@pyrello  All good! I attempted to run `yarn install` and `yarn build` within Icon Browser's `node_modules/uids` folder, which did succeed, but upon trying to use the components, I received this error:

```
app.js:1201 Uncaught TypeError: Cannot read properties of null (reading 'isCE')
    at renderSlot (runtime-core.esm-bundler.js?9feb:2947:1)
    at Proxy._sfc_render (uids.es.js?175d:1:1)
    at renderComponentRoot (runtime-core.esm-bundler.js?d2dd:895:1)
    at ReactiveEffect.componentUpdateFn [as fn] (runtime-core.esm-bundler.js?d2dd:5059:1)
    at ReactiveEffect.run (reactivity.esm-bundler.js?89dc:185:1)
    at setupRenderEffect (runtime-core.esm-bundler.js?d2dd:5185:1)
    at mountComponent (runtime-core.esm-bundler.js?d2dd:4968:1)
    at processComponent (runtime-core.esm-bundler.js?d2dd:4926:1)
    at patch (runtime-core.esm-bundler.js?d2dd:4518:1)
    at mountChildren (runtime-core.esm-bundler.js?d2dd:4714:1)
```

Which I Googled around for and found this post from: https://github.com/vuejs/core/issues/4344

> That's because you have two distinct copies of the Vue package being used, one in each package.
> 
> Vue uses a global singleton to track the current rendering instance - having more than one copy included will inevitably lead to such issues.

So I assumed I somehow duplicated or otherwise blew up my local UIDS components/package by attempting to run `yarn build` inside of `node_modules`, but I'm not sure.

quamsta:
Yeah ignore that above comment-- there was something wrong locally with my environment.

bspeare:
# Changes that need to be documented

- The major change in this that we need capture in our changelog is that buttons using `bttn bttn--outline bttn--tertiary` like on https://tippie.uiowa.edu/kira-price-story-undergraduate will render with a white background instead of being transparent.   The button classes would need to be changed to `bttn bttn--transparent` like on https://tippie.uiowa.edu/iowa-mba/curriculum/dual-degrees where the sections alternate between background colors. 
- Additionally, we will need to change the default card button classes to `bttn bttn--transparent bttn--sans-serif`. 
- `bttn__apply` has been removed.
- `bttn__link` has been removed `bttn borderless bttn--transparent bttn--sans-serif` should be used as a replacement.

bspeare:
Can we figure out a way to test these buttons with multiple background classes in this pr?  I've been working with https://uids.brand.uiowa.edu/branches/feature_button_tertiary_fixes/components/preview/kitchen-sink.html to test, but it is not ideal.

bspeare:
Should we add https://github.com/uiowa/uids/issues/714 to `_border.scss` here?

bspeare:
Did we also want to remove the BEM nesting on https://github.com/uiowa/uids/blob/2a8460641aaf054fb23cdbe04e1d231de74f422c/src/components/button/button.scss?

pyrello:
> `bttn--apply` has been removed. It is only used in viewbooks and perhaps we should move it to https://github.com/uiowa/vuejs-site-template?

What is `bttn--apply` doing?

bspeare:
I thought it was used in the viewbooks but I guess not.   I would guess I added it for a `uiowa.edu` header button that is no longer in use.

GaryRidgway:
Screen Shot 2022-06-30 at 10 08 47 AM
it is used on viewbooks to add the proper icon to an apply button
this is live on the admissions viewbook

GaryRidgway:
it serves the same functionality as `bttn--visit`, `bttn--ask`, and `bttn--connect`

bspeare:
@GaryRidgway I updated the class name in https://github.com/uiowa/uids/pull/716#issuecomment-1170505678.  It is a different class than the viewbook example.

bspeare:
I removed the BEM nesting and also removed the `bttn--link` modifier and added a entry for it in https://github.com/uiowa/uids/pull/716#issuecomment-1170505678.

GaryRidgway:
Future work:

- remove medium option
- add the option to align the icon left.
- remove serif option

bspeare:
Should we remove this `serif` option?

Screen Shot 2022-06-30 at 1 11 17 PM

quamsta (dismissed):
Hey all, these buttons are looking pretty fresh. I was able to use the latest commit from this branch in the Icon Browser and tested several variations and they're all looking pretty nice, landed on tertiary, small:

Screen Shot 2022-07-05 at 11 03 50 AM

Screen Shot 2022-07-05 at 11 04 01 AM

I'm going to approve this since it looks great on my end, but anyone should feel free to add their own review/approval here before this gets merged. Thanks everyone!

pyrello:
> Should we remove this `serif` option?

I thought we specifically decided to add that. Do you want to not have it anymore?

bspeare:
@pyrello I don't see us using the serif font any time soon.   Maybe we could shelve it for now?

quamsta (approved):
Approving this because apparently my old review got stale!

PR #725, opened 2022-07-13
Transparent link hotfix

GaryRidgway:
currently when you are hoving a transparent full with button with no border it looks like this:

Screen Shot 2022-07-13 at 10 50 08 AM

With this hotfix, it looks like this:

Screen Shot 2022-07-13 at 10 51 11 AM

### DIFF
--- a/src/components/button/button.scss
+++ b/src/components/button/button.scss
@@ -109,7 +109,7 @@
   }
 
   &.bttn--transparent:after {
-    position: static;
+    display: none;
   }
 }
 


pyrello:
Why does the button move to the center of the canvas?

bspeare (approved):
Good fix!

bspeare:
@pyrello the full width option stretches across the canvas and makes it appear in the center in that screenshot.

Issue #705, opened 2022-06-02
tables--gray-borders class doesn't have top border


bewoldt:
# Problem
If you add the class `table--gray-borders` to a table to add the gray borders, there is no top border. 

<img width="875" alt="Screen Shot 2022-06-02 at 11 41 37 AM" src="https://user-images.githubusercontent.com/731952/171681340-88f809ce-e3bc-48b9-8286-2af027524473.png">

joewhitsitt:
is this only because the table doesn't have a table header?

bewoldt:
> is this only because the table doesn't have a table header?

Yes, the top border is missing when you do not specify a table header row. But, if you were to choose a column as your header, this would still be an issue.

bspeare:
Closed through https://github.com/uiowa/uids/pull/735.

PR #735, opened 2022-07-27
Feature table updates

bewoldt:
Fix for #705 

Adds a border top back when using the table--gray-borders class. The only issue I've seen with this is when you have a caption that has the class "element-invisible" the top border doubles up. I don't see this as an issue since it is impossible for a user to add this class to a caption in the wysiwyg editor.

I also set a default line-height for tables to be the default body line-height which is 1.15. I noticed that tables created in the main body content area had a line-height of 1.7 but tables created within a component like the Text Area or Collection have 1.15. The main body content table are getting a line-height set by a class ".text-formatted div" of 1.7. I'm open to ideas on what the default should be but I set it as 1.15 for now.

### DIFF
--- a/src/components/tables/tables.scss
+++ b/src/components/tables/tables.scss
@@ -218,5 +218,4 @@ table.table--gray-borders {
     }
   }
 
-    th, td {
 }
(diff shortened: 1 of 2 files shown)


bewoldt:
sandbox page with some tables. https://sandbox.prod.drupal.uiowa.edu/more-tables

bspeare (approved):
Nice fix, thanks @bewoldt!

PR #759, opened 2022-08-18
4.x - CTA

### LINKED ISSUE #782: 4.x - Finish CTA component
Finish the work started in https://github.com/uiowa/uids/pull/759.

### LINKED ISSUE #897: [Bug] Safari - Antonio font is not being loaded
screenshot_2023-11-02_at_10 29 19_am

Button does not appear to be using Antonio when being viewed in Safari.

bspeare:
Resolves https://github.com/uiowa/uids/issues/782. 
Resolves https://github.com/uiowa/uids/issues/897.


src/components/cta/Cta.stories.js
```diff
@@ -0,0 +1,119 @@
+import UidsCta from './Cta.vue';
+import UidsHeadline from '../headline/Headline.vue';
```
pyrello: It looks like this is not being used in here?

src/components/cta/cta.css
```diff
@@ -0,0 +1,253 @@
+h1,
```
pyrello: I think we don't typically include this file. It gets generated or is not used. Either way, it will probably change as we start to add custom elements.

src/components/cta/Cta.stories.js
```diff
@@ -0,0 +1,119 @@
...
+    orientation: {
+      name: 'Orientation',
+      options: ['', 'left', 'left-right', 'center'],
+      control: {
+        type: 'select',
+        labels: {
+          '': 'Centered (default)',
+          'left': 'Left aligned',
+          'left-right': 'Left aligned with button on the right',
+          'center': 'Center',
```
pyrello: I'm confused by the Centered vs Center option. It appears that the Centered option is stacked, vs in a row (similar to cards). It seems like what the property is encapsulating might need to be split out into two things? Or tweaked a bit?

pyrello:
I guess I had missed that this was in progress and was so far along. Nice work on this so far. I think we need to refine how we are thinking about the variations (what is changing) and possibly split the properties to reflect that, but otherwise this is looking really good.

src/components/cta/Cta.stories.js
bspeare: Thanks for the feedback on this @pyrello .  I agree, what do you think about this? https://github.com/uiowa/uids/pull/759/commits/e4136c4648107e238f7784640d8b6267e9dfaa74.  The "Left aligned with button on the right" is now just a button alignment option in this latest version. 


Screenshot 2023-11-03 at 12 08 54 PM

pyrello: I think this does help clarify it a bit. I'd like to play with the controls a bit more.

GaryRidgway (dismissed):
After these small tweaks, it looks good!

pyrello:
Some thoughts for next steps after this:
- Implementing properties (`headline_style`) on components (CTA, Card) for nested components (Headline) doesn't scale very well. That is, it introduces a lot of redundancy because each component that implements a property to pass through to the nested component has to have essentially the same property as the nested component. For the headline example, this means that every component that implements the headline component needs to have a `headline_text_style`, `headline_underline`, and `headline_highlight` property that is the same as the relevant properties on the headline component. If we make changes to the logic of how the headline component works, we have to propagate those changes throughout all the parent component properties as well.
- I think we can resolve this by doing something like what we did with backgrounds, where we created a shared resource that can be pulled into a component. If there are logical changes, then we update them in the shared resource and they are propagated to every place where they are implemented.
- We'll probably need a way to only pull in the properties we need for a given component. There may be valid reasons why we want to hardcode/restrict some options in a nested component, so we don't want to have to accept them as all or nothing.

pyrello (dismissed):
I added some comments about future updates, but I think this is good to get this component added.

GaryRidgway (approved):
BLAM

PR #845, opened 2023-01-06
Callout component 4.x

### LINKED ISSUE #844: 4.x - Create callout component
As a UIDS user, I can insert a callout component into the body of my text area 

<details>
<summary>Original issue description</summary>
I don't think we want to call this a pullquote since they are meant to duplicate content from the body: https://component.gallery/components/quote/#pull-quotes.  I could see us developing a pullquote variant of the blockquote that is a larger font size in the future. 

I see this as essentially an inline text area box where all WYSIWG styles can be used, the most important aspect of the callout is the container padding, the width sizing, and the left/right alignment within body copy. 

Callouts and asides examples:
- https://designsystem.ontario.ca/components/detail/callouts-asides.html
- https://github.com/uiowa/uids/issues/9

Styles
- All styles are set based on container classes (no BEM elements, just modifiers)

Size (width)
- sm
- md
- lg

Alignment
- left align callout
- right align callout

Additional style options
- Borders
- Background
- Maybe a general sizing increase or decrease of all elements inside the callout to set visual contrast against the body copy. 
</details>

### Updates that need to be made to our inline sizing solution
- [x] Single image in column.  
    - We have determined this to be a content error, and an edge case that our previous system lacked in covering.
- [x] Add `<figure>` example and refactor accordingly.  
- [x] Add image example and refactor accordingly.
- [x] Add video example and refactor accordingly.

<details>
<summary>Details of current issues</summary>  

## Single image in WYSIWYG:
- Currently, when there is only one item in the column, it does not span the full width, as it is trying to size to the desired width with the WYSIWYG. For example, there is only one thing in a WYSIWYG, an image, and it is sized small. 

Example:

## Potential solution:

The following addition to the stylesheet will allow these styles to inherit the previous sizing before the container queries, but it is likely this is not a complete solution.
```@supports (contain: inline-size) {
  *[class*="inline--size-"] {
    width: 100%;

    &:only-child {
      width: inherit;
      margin: 0;
    }
  }
}
```

We should have a discussion on how we want to style `:only-child` images.

## View mode styling:

Currently, there are issues with the SNV2 view modes being sized in the column containers. These do serve a certain purpose in the sense that they can have a static size at lower image sizes, but additionally they are not covered under our current styling scheme. 

Here are some examples:

Small view mode:

Large view mode:

Small - no crop:

## Potential solution:
Write custom styles to specifically handle these view modes
### Or
Write an update hook to remove these view modes from SNV2&3 and have a conversation on what the path forward is for updating them. 

## Figure inline sizing:
- Currently, when a caption is displayed with an image, the two are wrapped in a `<figure>`. This breaks the styles for inline sizing.

Example: 

## Potential solution:
Re-define how the images are styled in uids, keeping figures in consideration. Then, propagate necessary changes to uids_base.

</details>

bspeare:
Resolves https://github.com/uiowa/uids/issues/844


src/components/callout/Callout.vue
```diff
@@ -0,0 +1,50 @@
...
+  },
+  alignment: {
+    type: String,
+    default: 'right',
+    validator: function (value) {
+      return ['', 'left', 'right', 'center'].indexOf(value) !== -1;
+    },
+  },
+  ...Background.props,
+  ...Media.props,
```
pyrello, line 14: Is there a reason this is being included? I can't see that it is actually being used.

bspeare: Background is being used, but not media.

GaryRidgway: Media props have been removed

pyrello:
I think we are going to leave this in draft to keep it open in case there are additional changes needed for uiowa/uiowa#6074.

src/components/size/size.scss
```diff
@@ -0,0 +1,133 @@
+@import '../../assets/scss/variables';
+@import '../../assets/scss/utilities';
+
+:root {
+  --element-spacing-lg: var(--space-lg-width-gutter);
+  --element-spacing-sm: var(--space-sm-width-gutter);
+}
+
```
bspeare: Instead of `:root` We might want to add a base `element--alignment` and `element--size` classes here to define the CSS variables on for each element.

GaryRidgway: We would be defining the variables for size and alignment though right? we wouldn't be re-defining variables per usage? Sorry, just trying to understand.

GaryRidgway: ex:

```
*[class*="element--alignment"], *[*[class*="element--size"] {  

    --element-spacing-lg: var(--space-lg-width-gutter);  
    --element-spacing-sm: var(--space-sm-width-gutter);
}
```
?

bspeare: I think @pyrello mentioned that we might want to move these to an `inline.scss` namespace?    I'm thinking about locally scoping the variables for just this component as discussed here https://blog.hubspot.com/website/css-variables.    We are using the global variables but setting their namespace at the component level `--element-spacing-sm: var(--space-sm-width-gutter);`.

bspeare: @GaryRidgway Yes, that's what I'm thinking but ideally would target these through BEM like `element-inline element-inline--size` and `element-inline element-inline--alignment` or something like that.

GaryRidgway:
My most recent commit should find things that are only children and inherit their sizing to the previous sizing before the column styles. This should cover the edge cases for there only being an image in a text area. If we do not want it, please feel free to remove it.

pyrello:
> My most recent commit should find things that are only children and inherit their sizing to the previous sizing before the column styles. This should cover the edge cases for there only being an image in a text area. If we do not want it, please feel free to remove it.

Does this solve a problem in UIDS? Or is it a SiteNow thing? If it solves a problem for UIDS, then I would like to see something in the inline storybook story to show what it is doing. However, it sounds like maybe this is a layout builder edge case, @bspeare?

GaryRidgway:
It is an edge case that could occur in uids as well, but you are correct, I didn't make a story for it. While we are reassessing, we can decide if it stays or not or if we need to make a story for it.

src/components/inline/inline.scss
```diff
@@ -86,8 +86,35 @@
...
+
+  @container column (width < #{$break-xxsm}) {
+    .inline--size-small {
+      &.inline--align-left,
+      &.inline--align-right, {
+        margin-inline-start: 0;
+        margin-inline-end: 0;
+      }
+    }
+  }
```
pyrello: In a narrow-first paradigm, anything less than our smallest breakpoint is the base case, so this should not be wrapped with a `@container` constraint.

GaryRidgway:
I had a neat idea to try and make a function to parse our column classes to output labels, but the @pyrello  on my shoulder told me it may have been too much.[ I have enshrined it in this compare statement so we can look at what it would do if we ever wanted to use it in the future!](https://github.com/uiowa/uids/compare/2a7a8bc...0cbcd9c)

GaryRidgway:
On first glance, the videos really don't want to grow in height...

GaryRidgway:
~Seems like we need to adjust to allow remote videos and whatnot not be mapped. Blargh.~

Not sure what keeps happening, but I am obviously having a hard time keeping everything in this PR straight, lol. Don't think this is an issue.

src/components/media/media.scss
```diff
@@ -86,6 +86,13 @@
   }
 }
 
+[class*='media--'] {
```
pyrello: Could this just be `.media`?

GaryRidgway: [It can!](https://github.com/uiowa/uids/pull/845/commits/1d5e9eb7894f4020f28938adb07147aefc370d90)

src/assets/scss/_variables.scss
```diff
@@ -98,6 +100,25 @@ $xlg: 2.5rem; // 40px
...
+$break-lg: 1200px;
+$break-page-container: 84.375em;
+$break-xlg: 106em;
 
 // Drop shadow.
+
 $soft-drop-shadow: 0px 1px 2px rgba(0, 0, 0, 0.4);
+
+
+:root {
```
bspeare: We should probably move these to `base.scss` so they don't load on every component - correct?

pyrello (approved):
Looks good to merge.

PR #852, opened 2023-01-10
Added column wrappers around grid slots and added container query name

bspeare:


src/components/grid/grid.scss
```diff
@@ -18,3 +18,12 @@
     }
   }
 }
+
+// @todo move this so a file that only storybook uses?
```
bspeare:
```suggestion
// @todo move this to a file that only storybook uses?
```

src/components/grid/grid.scss
```diff
@@ -18,3 +18,12 @@
     }
   }
 }
+
+// @todo move this to a file that only storybook uses?
```
bspeare:
```suggestion
```

pyrello:
I was trying to get something like this working: https://michaelnthiessen.com/supercharge-your-slots/#adding-slots-dynamically. I wasn't able to figure it out and it isn't worth spending more time on for now, but I want to come back to it eventually.

pyrello (approved):
This should be good to get us started.

PR #853, opened 2023-01-20
Pulled changes over from #845 by @GaryRidgway.

pyrello:
# To test
```
yarn storybook
```
- Verify no deprecation warnings.

### DIFF
--- a/src/assets/scss/_utilities.scss
+++ b/src/assets/scss/_utilities.scss
@@ -486,7 +486,7 @@
   // Make sure pixels are expressed in rems for min.
   @if (type-of($min) == "number") and ($min != 0) {
     @if (unit($min)  == 'px') {
-      $sans_unit_min: math.div($min, 1px);
+      $sans_unit_min: $min / 1px;
       $min: math.div($sans_unit_min, ($remsize + 0rem));
     }
   }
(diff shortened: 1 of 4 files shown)


GaryRidgway:
Sheesh, this was stupid. I had to remove divides as well because they were depreciated in whatever version of sass that Sitenow was using. So i just wrote my own.

src/assets/scss/_utilities.scss
```diff
@@ -1,3 +1,5 @@
+@use "sass:math";
```
bspeare: Do we still need this line?

bspeare:
Let's make sure we update these once it is moved to uids_base: https://github.com/uiowa/uiowa/search?q=containerClamp.

src/assets/scss/_utilities.scss
pyrello:
```suggestion
```
I don't think so.

src/assets/scss/_utilities.scss
```diff
@@ -1,3 +1,4 @@
+
```
pyrello:
```suggestion
```

bspeare (approved):
These look like the expected values.   Code reviewed, not tested.  Let's not forget https://github.com/uiowa/uids/pull/853#issuecomment-1400578987 and also updating the 3.x branch?

PR #857, opened 2023-01-30
Make the click-a11y script callable

### LINKED ISSUE #856: Make click-a11y script callable by external script.
# Problem

As discovered in https://github.com/uiowa/uiowa/issues/6138, the click-a11y script we use for cards and banners is not written in a way that allows it to be called by another script. It is loaded and runs automatically right away. This doesn't work for AJAX though because the script never re-runs after it has first been loaded. To work with AJAX, the script should be callable so it can be triggered to run on AJAX responses as well.

# Proposed solution

Refactor the click-a11y script to allow it to be called by another script.

pyrello:
Resolves #856.

Also, as part of this work, the headline component was updated to not include properties related to the content containing a link or not. Since the inclusion of the link doesn't have any affect on how the headline gets styled, we can just let the headline component be a wrapper around whatever text or link gets inserted into and avoid having the extra logic related to applying properties based on the link.

# To test

Using https://uids.brand.uiowa.edu/branches/callable_click_a11y/ to test:
- Check card examples against https://uids.brand.uiowa.edu/latest/ to see if there are any regressions.
- Confirm that unlinked card is not linked: https://uids.brand.uiowa.edu/branches/callable_click_a11y/?path=/story/components-card--with-no-link

### DIFF
--- a/src/components/card/Card.vue
+++ b/src/components/card/Card.vue
@@ -212,7 +212,7 @@ onMounted(() => {
 
     <div class="card__body">
       <header v-if="$slots.title">
-        <uids-headline :text_style="headline_style" class='test'>
+        <uids-headline :text_style="headline_style">
           <!-- @slot The title of the card. HTML is allowed. -->
           <a v-if="headlineLink" :href="headlineLink" class="click-target">
             <slot name="title">Title</slot>
--- a/src/components/card/Card.vue
+++ b/src/components/card/Card.vue
@@ -1,5 +1,5 @@
 <script setup lang="ts">
-import { computed, useSlots } from 'vue'
+import { computed, onMounted, useSlots } from 'vue';
 import { className } from '../utlity'
 import UidsHeadline from '../headline/Headline.vue'
 import UidsButton from '../button/Button.vue'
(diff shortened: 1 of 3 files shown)


src/components/card/Card.vue
```diff
@@ -185,23 +197,35 @@ const detailsElement = computed(() => {
...
+          name="media"
+          v-else
+        ></slot>
       </div>
     </div>
 
     <div class="card__body">
       <header v-if="$slots.title">
-        <uids-headline :url="headlineLink" :text_style="headline_style">
+        <uids-headline :text_style="headline_style" class='test'>
```
joewhitsitt: is this still needed? `class='test'`

pyrello:
```suggestion
        <uids-headline :text_style="headline_style">
```

pyrello:
Updated summary to reflect the changes to the headline component. Shouldn't affect testing instructions, since the headline is included in the card and will get tested as part of testing the cards.

joewhitsitt (approved):
From the testing instructions, this seems fine.

PR #874, opened 2023-03-14
Some inline size refactoring

pyrello:
Makes the base case (no container queries) more robust by adding more breakpoints and moving some logic from the container query section into the base case.

# To test

This updates the callout alignment, sizing, and margins so that it works without container query code. 
- Test callouts with breakpoints on https://uids.brand.uiowa.edu/latest/?path=/story/elements-inline--three-column and that the callout stretches to 100% when columns get squished. 
- Test with https://github.com/uiowa/uids/blob/4.x/src/components/grid/grid.scss#L66-L72 commented out and verify that sizing and alignment in https://uids.brand.uiowa.edu/latest/?path=/story/elements-inline--narrow still works.

### DIFF
--- a/src/assets/scss/_breakpoints.scss
+++ b/src/assets/scss/_breakpoints.scss
@@ -2,7 +2,13 @@
 
 // Media Queries ///////////////
 @mixin breakpoint($class) {
-  @if $class==xsm {
+  @if $class==xxsm {
+    @media (min-width: $break-xxsm) {
+      @content;
+    }
+  }
+
+  @else if $class==xsm {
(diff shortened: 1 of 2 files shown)


GaryRidgway:
All the testing instructions work as stated, but I am seeing this on three columns when commenting out the lines in instruction two:
Screenshot 2023-03-16 at 9 59 29 AM
Is that something we should investigate, or are we aware of this problem and not worried?

GaryRidgway (approved):
Based on the testing instructions, this looks good!

PR #875, opened 2023-03-23
Add clear:left/right to inline--align-left/right

cory-skeers:


bspeare (approved):
Tested this locally and it appears to work as expected.

cory-skeers:
Thanks @bspeare!

PR #829, opened 2022-12-22
Fixing antonio weights for browser consistency 

bspeare:
Resolves https://github.com/uiowa/uids/issues/772. 

This fixes the font-weight issue we were having with safari without having to switch to Google fonts.  Also, with upcoming privacy concerns I think it makes sense to continue to locally host our fonts. 

There is an update pr with a more updated Antonio.woff font in https://github.com/uiowa/uids/pull/828, however that update requires a lot of spacing adjustments that we can avoid but sticking with our older version of Antonio here. 

# How to test

In Safari, compare https://uids.brand.uiowa.edu/branches/feature_fix_antonio_weight/components/detail/stat--default.html with https://uids.brand.uiowa.edu/components/detail/stat--default.html.   The stat font should look thin on uids.brand.uiowa.edu and the feature branch font weight should look thicker.  

Compare https://uids.brand.uiowa.edu/branches/feature_fix_antonio_weight/components/detail/stat--default.html  in Safari and Chrome and verify that the font weights look the same.

PR #877, opened 2023-04-13
Feature stat update prefix

bspeare:
Updated spacing for prefix/suffix after https://github.com/uiowa/uids/pull/829 adjusted the font slightly.


pyrello (approved):
Looks good.

PR #880, opened 2023-05-04
Which container class should we use for uids and uids_base?

bspeare:


bspeare:
Closing in favor of https://github.com/uiowa/uids/pull/922.

PR #881, opened 2023-06-12
Updated nvm config to use Node 16

pyrello:
# To test

```
nvm install && nvm use && yarn install && yarn storybook
```


joewhitsitt:
we are now running node 18 here https://github.com/uiowa/uids/blob/4.x/.nvmrc

PR #884, opened 2023-06-23
Removed card title class name that is no longer used

bspeare:
# How to test

This fixes a `bg-black` and `bd-dark` mixin that was targeting `card__title`.   Since we no longer use that class, I am targeting the `headline.default` class for the mixin. 

- Code review only


bspeare:
Closing this.

PR #886, opened 2023-07-14
Toggle nav updates

bspeare:
This pr updates the `toggle-nav.js` to set the off-canvas menu positioning instead of trying to do this with CSS.   In addition, there's also a `MutationObserver` observer added to watch for dynamic changes to the header position.


pyrello:
So many double quotes...

PR #891, opened 2023-08-23
Alerts: updates for uids_base integration

bspeare:
Work for https://github.com/uiowa/uiowa/pull/6810

# How to test

- Review the `alert--icon` modifier class that is providing the `display: flex` property to the alert that allows the icon and alert content to display the side by side.  This modifier was added because when alerts are added in the SiteNow WYSIWYG they are not enclosed in a `<div>` wrapper and the `display: flex` that was previously added by default to the `.alert`class rendered the alert text as unreadable. 

Basically, this format of alert below needs to work:
```
<div class="alert alert--info">Alert text</div>
```

### DIFF
--- a/src/components/alert/Alert.stories.js
+++ b/src/components/alert/Alert.stories.js
@@ -26,6 +26,12 @@ export default {
         category: 'Display options',
       },
     },
+    icon: {
+      name: 'Icon',
+      table: {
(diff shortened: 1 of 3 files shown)


pyrello (approved):
Reviewed, not tested.

Issue #895, opened 2023-10-17
Stub out components not yet in 4.x


briand44:
3.x currently has components not yet in 4.x. We should stub those out in 4.x prior to https://github.com/uiowa/uids/issues/892
```[tasklist]
### Tasks
- [x] Create a list of components that 3.x has that 4.x does not have
- [x] Use that list to create stubs in the 4.x branch.
```

GaryRidgway:
**These are the components we have identified from v3 that should move**

Accordion
Background
Badge
Banner
Blockquote
~~Colors~~
CTA
Footer
Slider
Stat
Tables
Tabs

pyrello:
Looks like this wasn't set to automatically close with #896

PR #896, opened 2023-10-30
Stub out unimplemented uids3 components in uids4

GaryRidgway:
Tracks: https://github.com/uiowa/uids/issues/895

## To test
1. pull down this branch of UIDS
2. Check that the following items are stubbed out in the "Not implemented" section
Accordion
Background
Badge
Banner
Blockquote
~~Colors~~
CTA
Footer
Slider
Stat
Tables
Tabs


joewhitsitt:
I can try to pull and test this

joewhitsitt:
Not in-scope but I noticed we have story/basic-example--default available. Should that be removed?

/story/components-stub--info is also showing up. not sure if that was intended based on this PR's summary.

I missed the portion of EoD when this was presented and the conversation after it, but do we want to say, "We are in the process of moving this component to UIDS 4," or should we switch to a less committed, "This component has yet to be implemented in UIDS 4"

joewhitsitt:
There is a "stub" for every component listed above.

GaryRidgway:
> Not in-scope but I noticed we have story/basic-example--default available. Should that be removed?

I think so, if we need it, we can always add it back from an older commit.

> /story/components-stub--info is also showing up. not sure if that was intended based on this PR's summary.

Good catch, I have removed that.

> I missed the portion of EoD when this was presented and the conversation after it, but do we want to say, "We are in the process of moving this component to UIDS 4," or should we switch to a less committed, "This component has yet to be implemented in UIDS 4"

I like your non committal option! I will add that.

pyrello:
> > I missed the portion of EoD when this was presented and the conversation after it, but do we want to say, "We are in the process of moving this component to UIDS 4," or should we switch to a less committed, "This component has yet to be implemented in UIDS 4"

> I like your non committal option! I will add that.

I think we had previously discussed a version of this that included the word "yet" and thought that maybe it conveyed that it was taking a while. The removal of "yet" was meant to make the statement neutral with regard to whether we should have moved it already or not.

GaryRidgway:
> > > I missed the portion of EoD when this was presented and the conversation after it, but do we want to say, "We are in the process of moving this component to UIDS 4," or should we switch to a less committed, "This component has yet to be implemented in UIDS 4"
> 
> > I like your non committal option! I will add that.
> 
> I think we had previously discussed a version of this that included the word "yet" and thought that maybe it conveyed that it was taking a while. The removal of "yet" was meant to make the statement neutral with regard to whether we should have moved it already or not.

Ah, good point. I've adjusted it again.

GaryRidgway:
> Not in-scope but I noticed we have story/basic-example--default available. Should that be removed?


It has now been removed!

joewhitsitt:
The recent changes have satisfied my concerns. It looks like @pyrello is also taking a look so I am going to hold on approval.

pyrello (approved):
This is a good step towards what we need to move to 4 as the default. Nice work, @GaryRidgway!

Issue #892, opened 2023-10-05
4.x - Move 4.x branch to replace 3.x at uids.brand.uiowa.edu


bspeare:
As a campus developer, I can visit uids.brand.uiowa.edu and see an updated version of UIDS4 using Storybook and Vue.js

# Proposed Solution

```[tasklist]
### Tasks
- [x] Update build process to move 4.x branch to replace 3.x at uids.brand.uiowa.edu
- [x] Move 3.x branch so that is available at an archived directory
- [x] Hide unfinished components in 4.x
- [x] Build pages in 4.x branch to link to 3.x branch components that aren't finished in 4.x yet
- [x] Update links on brand.uiowa.edu to UIDS header and footer
- [x] Manually clean up gh-pages
```

joewhitsitt:
Just recently, I know folks have been referencing the header and footer "components". Do we need to stub these out as well?

https://brand.uiowa.edu/applying-brand/web

briand44:
Grooming conversation... Do minimum, stubs for things not transitioned and link to v3. Easy components to transfer likely address in a follow up.

GaryRidgway:
- Change v4 to be the main design system branch, and put v3 on a feature branch that will be accessed at uids.brand.uiowa.edu/v3
- Create stubs for 4x for components that we don't have components for in 4x.
- Link those stubs to the 3x versions of the components.

briand44:
Discussed having a separate slice for stubbing out components that haven't transitioned to 4.x yet and tackle that prior to this slice.

briand44:
Grooming conversation... Possibly a slice of hiding components in 3.x that we don't plan on migrating. Can't remove because still in use by uids_base. What does hiding entail? Removing documentation, remove/make inaccessible to fractal build process but leave css/scss.

Remove or make inaccessible to fractal build documentation files or components that we want to hide. We need to start by figuring out what the list is.

size - small to medium

bspeare:
Update GitHub actions build process for 3.x and 4.x.

e-marie-w:
Yesterday, we updated the build process, so uids.brand.uiowa.edu now reflects the 4.x branch. By clicking on a Not Implemented component, users can get to the 3.x components still. Knowing that, do we still need to create uids.brand.uiowa.edu/v3? Or is that what's meant by the `Move 3.x branch so that is available at an archived directory` step?

[...]

PR #905, opened 2023-11-17
3.x Fix build process so 3.x branches successfully build to 3.x directory

e-marie-w:
Yet another piece of #892. 

Removed code previously added to stop 3.x from building to /latest, in case it is also blocking us from building to /3.x.


pyrello (approved):
Reviewed, will be tested by merge + build. Let's see if it works!

PR #906, opened 2023-11-20
Cleanup gh-pages

e-marie-w:
Another piece of #892. 

Removed /components because /3.x has replaced it.
Removed /docs/v4.0.0 because /docs/v4.0.0-alpha1 has replaced it 
Removed /mono-theme and /themes/mandelbrot because they are not in use.
Removed subdirectories from /assets because they are now in /3.x/assets
Removed .html files that are now found in /3.x


pyrello (approved):
Reviewed. Let's see how it builds.

Issue #908, opened 2023-11-29
4.x - Provide base CSS stylesheets


bspeare:
## Story 

As a user of uids.brand.uiowa.edu, I can copy and paste markup and CSS from each component to use on my website or application. 

# 3.x Solution
For 3.x we have been providing CSS on https://brand.uiowa.edu/applying-brand/web#heading6951:
```
To utilize the brand bar on your website or application, you will need to use the following CSS files in the order that they are listed below.

    https://github.com/uiowa/uids/blob/main/dist/css/fonts.css
    https://github.com/uiowa/uids/blob/main/dist/css/reset.css
    https://github.com/uiowa/uids/blob/main/dist/css/components/iowa-bar.css

```

# Proposed  4.x Solution
- [ ] Define a build process to produce the assets defined above from 3.x solution
- [ ] Build it!!!

bspeare:
This might be blocked until https://github.com/uiowa/uids/issues/900 is complete.

pyrello:
> Maybe https://storybook.js.org/addons/display-element-css is worth trying first?

I'm assuming that this is something that outputs CSS that can be viewed and copied. I would argue that displaying the CSS rather than providing links that we intend to keep in place over time is going to exacerbate the issue we have where people consume UIDS by copying/pasting rather than directly using the assets. I think we should be focusing on trying to get to a point where people can reliably consume the assets because anything else is a maintenance headache for the implementing app and a support headache for us.

briand44:
Future slice for all components can be done in consultation with partners so they can weigh in on what would be most valuable to them.

PR #915, opened 2023-12-15
Add a build process for generating CSS files and instructions for how to use them

bspeare:
Relates to https://github.com/uiowa/uids/issues/908. 

# How to test

- `yarn production`
- `npx http-server ./www`
- Go to the Introduction page and confirm that CSS files are listed using the version number from `package.json`. 
- Go to the Brand Bar docs page and confirm that the CSS  files are listed using the version number from `package.json`.


src/assets/scss/reset.scss
```diff
@@ -290,7 +290,7 @@ select {
  */
 button,
 html [type='button'],
-/* 1 */
+  /* 1 */
```
pyrello, line 293: What's this about?

src/components/Introduction.stories.mdx
```diff
@@ -1,8 +1,21 @@
 import {Meta} from '@storybook/blocks';
+import {version} from '/package.json';
```
pyrello:
```suggestion
import { version } from '/package.json';
```

src/assets/scss/reset.scss
bspeare: The old extra spacing that I didn't intend issue.

package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "uids",
   "type": "module",
-  "version": "4.0.0-alpha1",
+  "version": "4.0.0-alpha2",
```
pyrello: To actually generate the assets that will be linked in the docs, we have to create a new release. If we are going to create a new release, we need to update the version, since we already have a release for [`4.0.0-alpha2`](https://github.com/uiowa/uids/releases/tag/v4.0.0-alpha2)
```suggestion
  "version": "4.0.0-alpha3",
```

pyrello:
I'm going to look at this again today.

package.json
```diff
@@ -26,7 +26,8 @@
     "typecheck": "vue-tsc --noEmit -p tsconfig.vitest.json --composite false",
     "lint": "eslint . --ext .vue,.js,.jsx,.cjs,.mjs,.ts,.tsx,.cts,.mts --fix --ignore-path .gitignore",
     "storybook": "storybook dev -p 6006",
-    "build-storybook": "storybook build"
+    "build-storybook": "storybook build",
+    "release": "vite build && storybook build"
```
pyrello: tl;dr: This change should not be necessary once the plugin file is working correctly.

@bspeare and I talked about this. The `release` command was added because just running `storybook-build` was not resulting in the files being placed in the desired location (`/dist`). For all intents and purposes it looked like Storybook was skipping the plugin. Running `vite build` would result in the files getting placed in the correct location. However, when looking into this further, it appears that [Storybook should be consuming the root level vite config](https://storybook.js.org/docs/builders/vite#configuration), which contains the plugin being added in this PR, so theoretically it should actually be running. After some more troubleshooting (adding `console.log()` statements to the plugin), I was able to determine that the plugin is running. The generated files are being placed in `/storybook-static`, which seems to be set contextually as the root when the command is run in a production context.

vite-css-component.js
```diff
@@ -0,0 +1,68 @@
...
+      // Exclude Vue SFC style blocks
+      if (id.endsWith('.scss') && !id.includes('?vue&type=style')) {
+        const isBaseFile = files_to_compile.some(file => id.endsWith(file));
+
+        if (isBaseFile) {
+          // Extract the base file name
+          const baseFileName = path.basename(id, '.scss');
+
+          // Define the output CSS path for base files
+          const outputCssPath = `dist/css/base/${baseFileName}.css`;
```
pyrello: This seems problematic because the path is inherently going to be relative to whatever location this script gets run from. I suspect this might be why the files get placed in `/storybook-static` when the `storybook-build` command gets run.

vite-css-component.js
```diff
@@ -0,0 +1,68 @@
...
+          return null;
+        } else {
+          // Extract the component name and path
+          const matches = id.match(/\/components\/([^/]+)\/([^/]+)\.scss/);
+          if (matches) {
+            const componentName = matches[2];
+            const componentPath = matches[1];
+
+            // Define the output CSS path based on the component structure
+            const outputCssPath = `dist/css/components/${componentPath}/${componentName}.css`;
```
pyrello: Same issue here.

.github/workflows/gh-pages.yml
```diff
@@ -82,7 +82,7 @@ jobs:
 #      - run: npm -v
 
       - run: yarn install --immutable
-      - run: BASE_URL=$target_dir$branch/ yarn build-storybook -o www
+      - run: BASE_URL=$target_dir$branch/ yarn release -o www
```
pyrello: This change is likely unnecessary. See comments below.

package.json
pyrello: Since our build is currently just using Storybook's process, we might just want to use the `dist` files created in `storybook-static`. If we need to move them, we could potentially add that to the workflow script.

pyrello (changes requested):
I think this needs to be reworked a little bit to ensure that we are not adding things that aren't actually necessary. This will hopefully prevent us from adding complication that we'll have to decode in the future.

pyrello (approved):
Multiple rounds of testing by @bspeare and myself. This looks like a good next step.

PR #917, opened 2023-12-21
WIP

rramsey:


rramsey:
- All components are exported into the ./dist folder
 - Individual components can be imported like this in a vue app:
   `  import "/node_modules/uids/banner/banner.umd.cjs";`
- Individual componsnets can be used like this:
  `<uids-banner>I am a banner</uids-banner>`
- running `npm run build` in a vue app that consumes the component __appears__ to work

**To Do**
- deploy a built consumer of the uids button to a server and confirm that it runs as expected
- find a way to replace importing the cjs file with a simpler `import { BannerElement } from 'uids/main';`

PR #919, opened 2024-01-03
Updated bttn mixin to target span tags

bspeare:


pyrello (approved):
Reviewed, not tested.

PR #921, opened 2024-01-31
Updated brand bar docs to use correct path

bspeare:
`yarn production`
`npx http-server ./www`


pyrello (approved):
Looks good. Thanks for putting this together.

PR #923, opened 2024-04-03
Card container query updates

bspeare:
# How to test

Switched remaining left/right media queries to use the new mixin.


pyrello (approved):
Reviewed, not tested.

PR #924, opened 2024-05-13
Refactoring to always have access to the UIowa colors.

GaryRidgway:
This should do nothing at all to UIDS, but allow the logo to always be gold, even when the primary colors change.


joewhitsitt (approved):
Reviewed, not tested.

PR #935, opened 2024-07-01
Date Card

### LINKED ISSUE #934: Implement day/month design into card component
As it looks in https://github.com/uiowa/uiowa/issues/2727#issuecomment-834403434, implement the day/month design into the card component. 

117458545-06999c80-af10-11eb-9c1e-e3e6afeda6b6

```[tasklist]
### Tasks
- [ ] Use the media slot?
- [ ] Create story showing how the design adapts to all variants
- [ ] Grab calendar day/month styles from https://github.com/uiowa/uiowa/blob/a8c39d4939dad7a33d451fba00bf18725b9e1e0a/docroot/sites/registrar.uiowa.edu/modules/registrar_core/css/academic_calendar.css#L101 to start with
```

GaryRidgway:
resolves https://github.com/uiowa/uids/issues/934

This PR adds in media dates [as seen before on registrar](https://registrar.uiowa.ddev.site/calendars-and-deadlines/academic-calendar). This should allow them to be a bit more flexible in the media slot.

Additionally, this PR includes a refactor of sizes, creating more variations for media to accomodate smaller items in the media slot.

## To test
1. Test Media date: http://localhost:6006/?path=/story/components-card--media-date using all combinations of backgrounds and borders mobile/desktop view.   
2. Test on Grid http://localhost:6006/?path=/story/components-card--grid 
  a. using `<div class="media--date"><span class="media--date__month">June</span><span class="media--date__day">9</span></div>` in media slot.  
  b. Test all Orientation variations in combination with Background Colors and Border boolean.   
 4. Test "no-crop" option on http://localhost:6006/?path=/story/components-card--media-date for images using `` in media slot - adjust size from small, medium, to large and confirm it resizes and works.

### DIFF
--- a/src/components/card/Card.stories.js
+++ b/src/components/card/Card.stories.js
@@ -410,3 +410,18 @@ Grid.args = {
   grid_type: 'threecol--33-34-33',
   record_count: 3,
 }
+
+export const DateHighlight = GridTemplate.bind({})
+DateHighlight.args = {
+  ...Default.args,
+  orientation: 'left',
+  media_size: 'small',
+  media_shape: 'circle',
+  media:
+    '<div class="date-highlight">\n' +
+    '  <span class="month">June</span>\n' +
+    '  <span class="day">9</span>\n' +
+    '</div>',
+  grid_type: 'onecol',
+  record_count: 1,
+}
--- a/src/components/card/Card.stories.js
+++ b/src/components/card/Card.stories.js
@@ -415,8 +415,8 @@ export const MediaDate = GridTemplate.bind({})
 MediaDate.args = {
   ...Default.args,
   orientation: 'left',
-  media_size: 'no-crop',
-  media_shape: 'circle',
+  media_size: 'small',
+  media_shape: 'no-crop',
   media:
     '<div class="media--date">\n' +
     '  <span class="media--date__month">June</span>\n' +
--- a/src/components/card/Card.stories.js
+++ b/src/components/card/Card.stories.js
@@ -415,7 +415,7 @@ export const MediaDate = GridTemplate.bind({})
 MediaDate.args = {
   ...Default.args,
   orientation: 'left',
-  media_size: 'small',
+  media_size: 'default',
   media_shape: 'no-crop',
   media:
     '<div class="media--date">\n' +
--- a/src/components/card/Card.stories.js
+++ b/src/components/card/Card.stories.js
@@ -415,7 +415,7 @@ export const DateHighlight = GridTemplate.bind({})
 DateHighlight.args = {
   ...Default.args,
   orientation: 'left',
-  media_size: 'small',
+  media_size: 'xx-small',
   media_shape: 'circle',
   media:
     '<div class="media-date">\n' +
--- a/src/components/card/Card.stories.js
+++ b/src/components/card/Card.stories.js
@@ -411,16 +411,16 @@ Grid.args = {
   record_count: 3,
 }
 
-export const MediaDate = GridTemplate.bind({})
-MediaDate.args = {
+export const DateHighlight = GridTemplate.bind({})
+DateHighlight.args = {
   ...Default.args,
   orientation: 'left',
-  media_size: 'xx-small',
+  media_size: 'small',
   media_shape: 'circle',
   media:
-    '<div class="media--date">\n' +
-    '  <span class="media--date__month">June</span>\n' +
-    '  <span class="media--date__day">9</span>\n' +
+    '<div class="date-highlight">\n' +
+    '  <span class="month">June</span>\n' +
+    '  <span class="day">9</span>\n' +
     '</div>',
   grid_type: 'onecol',
   record_count: 1,
--- a/src/components/card/Card.stories.js
+++ b/src/components/card/Card.stories.js
@@ -418,9 +418,9 @@ DateHighlight.args = {
   media_size: 'small',
   media_shape: 'circle',
   media:
-    '<div class="date-highlight">\n' +
-    '  <span class="month">June</span>\n' +
-    '  <span class="day">9</span>\n' +
+    '<div class="media-date">\n' +
+    '  <span class="media-date--month">June</span>\n' +
+    '  <span class="media-date--day">9</span>\n' +
     '</div>',
   grid_type: 'onecol',
   record_count: 1,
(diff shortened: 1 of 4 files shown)


src/scss/components/card.scss
```diff
@@ -116,6 +116,34 @@
       object-fit: cover;
     }
 
+    .media-date {
```
pyrello:
```suggestion
    .media--date {
```
To make this consistent with other classes we have that extend the base `.media` class.

src/scss/components/card.scss
```diff
@@ -116,6 +116,34 @@
...
+      color: #000;
+      background-color: #ffcd00;
+      text-align: center;
+      display: flex;
+      flex-direction: column;
+      align-items: center;
+      justify-content: center;
+      padding: 0.9rem 0.6rem 0.2rem 0.6rem;
+
+      .media-date--month {
```
pyrello:
```suggestion
      .media--date__month {
```
Should be underscores according to BEM since this is a part of a larger element.

src/scss/components/card.scss
```diff
@@ -317,30 +349,50 @@
...
+  .media--x-small {
+    max-width: 11.375rem;
+
     @include container-query() {
       max-width: 11.375rem;
     }
   }
 
-  .media--small.media--widescreen {
+  .media-small {
```
pyrello:
```suggestion
  .media--small {
```
Probably should be this?

src/components/card/Card.stories.js
```diff
@@ -410,3 +410,18 @@ Grid.args = {
   grid_type: 'threecol--33-34-33',
   record_count: 3,
 }
+
+export const DateHighlight = GridTemplate.bind({})
```
pyrello:
```suggestion
export const MediaDate = GridTemplate.bind({})
```
I don't think we should call this a highlight since it is confusing in the context of other places we use that term.

Also, should this be just extending one of the single card classes? Or do we just find the grid presentation to be more useful and maybe we should get rid of the single card classes?

src/scss/components/card.scss
GaryRidgway: Yes! Thank you

src/components/card/Card.stories.js
GaryRidgway: Truthfully I don't find much value in looking at the cards in the singles. I dont get a clear picture of the intended look, and it is not possible to correctly show a media left or right example from what I've seen.

I will commit the change to the name though!

joewhitsitt:
I can review this, but I am trying to understand the downstream effects of the sizing change. Does this mean we will need to update layout builder style defintions and/or run an update hook to update existing media sizing?

joewhitsitt:
I recall this [issue](https://github.com/uiowa/uiowa/issues/7195) where we considered different terms for the sizing instead of heading down the x- xx-... naming convention. Do we want to start to establish that here?

>thumbnail, tiny, and extra small

joewhitsitt:
I haven't worked with UIDS/Storybook in a while. Is it just me or is the Orientation setting for cards hit or miss between the different "stories." Also, do we need some wrangling if stacked medium/large is used?

Screenshot 2024-07-02 at 3 49 47 PM

Update: I guess icon medium/large has the same effect but the icon within fills the space where as this text size stays the same.

GaryRidgway:
> I can review this, but I am trying to understand the downstream effects of the sizing change. Does this mean we will need to update layout builder style defintions and/or run an update hook to update existing media sizing?

I was thinking we would have to, but I recall being told that it might be as easy as a find and replace as well. So I think it might take some investigation to know

> I recall this [issue](https://github.com/uiowa/uiowa/issues/7195) where we considered different terms for the sizing instead of heading down the x- xx-... naming convention. Do we want to start to establish that here?
> 
> > thumbnail, tiny, and extra small

This could work as well, I think asking opinions on size names will come up in my EOD later today

> I haven't worked with UIDS/Storybook in a while. Is it just me or is the Orientation setting for cards hit or miss between the different "stories." Also, do we need some wrangling if stacked medium/large is used?
> 
> Screenshot 2024-07-02 at 3 49 47 PM

Yeah, the initial stories dont offer much unless you crop the viewport, but maybe some wrangling would be good. However, thats how images work as well, so it might be intended.

joewhitsitt:
Sounds like we are going to discuss a bit more in-depth after stand-up tomorrow

bspeare:
Here's an example of the no crop version for the date and a vertical image: https://codepen.io/bspeare/pen/KKLOEBV

joewhitsitt:
this is what i fear:
Screenshot 2024-07-03 at 10 11 47 AM

_surely nobody will use a ridiculous aspect ratio image like this_ :face_palm:

joewhitsitt:
i can pull this down to test again after standup

joewhitsitt (dismissed):
Looks good. Thanks for making the shift.

joewhitsitt:
Should we merge this and create a release for the next slice for registrar?

pyrello:
I haven't pulled and tested yet, but reviewing the code changes, it seems that the date display is going to look different in the case of cards with gold backgrounds. Specifically, there will be no gold circle, just the black date text. I think ideally, we would have appropriate color/styling for the media date for each background color option.

@bspeare suggested we could potentially solve for this now by adding a black border around the circle in the case of a gold background.

We don't have to tackle this now if we are anxious to get this merged, as I think we are meeting the needs of the case we developed this for.

joewhitsitt:
Good point:
Screenshot 2024-07-05 at 10 24 39 AM

joewhitsitt:
Is this okay to push?

Screenshot 2024-07-05 at 10 58 00 AM

```diff
diff --git a/src/scss/components/card.scss b/src/scss/components/card.scss
index 4de59f566c..cdfd9f30a0 100755
--- a/src/scss/components/card.scss
+++ b/src/scss/components/card.scss
@@ -213,8 +213,8 @@
 
   .media--date {
     position: relative;
-    color: $uiowa-black;
-    background-color: $uiowa-gold;
+    color: $white;
+    background-color: $uiowa-black;
     padding-top: $gutter;
     text-align: center;
     width: 6rem;
@@ -241,6 +241,12 @@
       font-weight: $font-weight-medium-bold;
     }
   }
+  &:not([class*="bg--gold"]){
+    .media--date {
+      color: $uiowa-black;
+      background-color: $uiowa-gold;
+    }
+  }
 }
 ```

pyrello:
> Is this okay to push?
> 
> Screenshot 2024-07-05 at 10 58 00 AM

Can you add a code comment indicating that we would like to revisit this eventually?

joewhitsitt (approved):

PR #936, opened 2024-07-15
[Card] Adjusted styles for media--date

bspeare:
# How to test
- Code review
- Test http://localhost:6006/?path=/story/components-card--media-date with all background colors
  - Add `element-invisible` class to `<div class="media--date">` and spacing should remain the same, the media slot should not collapse.

### DIFF
--- a/src/scss/components/card.scss
+++ b/src/scss/components/card.scss
@@ -213,9 +213,8 @@
 
   .media--date {
     position: relative;
-    color: $white;
-    background-color: $uiowa-black;
     padding-top: $gutter;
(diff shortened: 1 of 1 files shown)


GaryRidgway (approved):
Tested and looks good to me!

Something to keep in mid for the future is that it seems like the thinness of roboto on the black background may make it slightly harder to read on higher resolution monitors. Perhaps we should adjust the `-webkit-font-smoothing` property for this?

bspeare:
@GaryRidgway good question on the font smoothing.   Feel free to push up an adjustment to the font smoothing on the `media--date`.

PR #939, opened 2024-10-01
[Card] margin adjustment for subtitle

bspeare:
This pr moves the negative margin to the card__subtitle so that when cards with title + meta are used, they won't be so close together.  

# How to test 
- On `/?path=/story/components-card--person-profile` confirm that subtitle spacing matches https://uids.brand.uiowa.edu/latest/?path=/story/components-card--person-profile.   
- Remove subtitle and add meta text and compare against https://uids.brand.uiowa.edu/latest/?path=/story/components-card--person-profile and confirm that there is a little extra spacing between title and meta fields.

### DIFF
--- a/src/scss/components/card.scss
+++ b/src/scss/components/card.scss
@@ -236,12 +236,13 @@
   font-size: 1.3rem;
   font-weight: $font-weight-medium;
   line-height: 1.4;
+  margin-top: -0.4rem;
 }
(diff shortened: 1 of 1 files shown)


joewhitsitt:
pulling to test

joewhitsitt (approved):
works as described

PR #941, opened 2024-11-05
[Card pre-title] Switched to opacity instead of directly assigning a color

bspeare:


joewhitsitt (approved):
Nice catch and clever solution

PR #942, opened 2024-11-07
[SASS] More deprecation fixes for Sass 3.0.0

bspeare:
# How to test

- `yarn install` 
- `yarn storybook`

### DIFF
--- a/src/scss/abstracts/_utilities.scss
+++ b/src/scss/abstracts/_utilities.scss
@@ -137,51 +137,47 @@
   font-size: $h2-font-size;
   line-height: 1.25;
   font-weight: $font-weight-medium;
+  // https://garyridgway.github.io/clampCalculator/?&minWidth=600&maxWidth=1310&minFont=1.95&maxFont=2.3
+  font-size: clamp(1.95rem, calc(0.7887vw + 1.6542rem), 2.3rem);
   @include margin($top: 0, $bottom: $xsm);
   &:not(:first-child) {
     @include margin($top: $lg);
   }
-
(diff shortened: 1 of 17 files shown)


src/scss/abstracts/_utilities.scss
```diff
@@ -137,51 +137,47 @@
   font-size: $h2-font-size;
   line-height: 1.25;
   font-weight: $font-weight-medium;
+  // https://garyridgway.github.io/clampCalculator/?&minWidth=600&maxWidth=1310&minFont=1.95&maxFont=2.3
+  font-size: clamp(1.95rem, calc(0.7887vw + 1.6542rem), 2.3rem);
```
pyrello, line 139: Should we get rid of the other `font-size` declaration?

src/scss/components/cta_uids3.scss
```diff
@@ -0,0 +1,121 @@
+@import '../abstracts/variables';
```
pyrello, line 1: Why does this file have `_uids3` in the name when other files being moved over do not?

bspeare: We need it there until https://github.com/uiowa/uiowa/issues/8234.  I'm fine with any other name, but there's already a newer `cta.scss`.

src/scss/abstracts/_utilities.scss
bspeare: The first `font-size` was there as a fallback for older browsers.  We can probably remove it now?

pyrello: No worries if it is a valid use case. Just noticed it appeared to be a duplicate.

pyrello (approved):
Looks good. Tested this via uiowa/uiowa#8144.

bspeare:
Thanks @pyrello!

PR #948, opened 2024-11-20
[Card] Icons

bspeare:
# How to test
```
yarn storybook
```

- On the grid story http://localhost:6006/?path=/story/components-card--grid, test out all possible variant combos that you can think of along with background colors. 
- Adjust to use some of the different icons added to this pr.


pyrello (approved):
Tested via uiowa/uiowa#8298.

PR #957, opened 2024-12-10
[Stat] Hover fix

### LINKED ISSUE #947: Stat component transition causes abrupt change in parent component height
# Motivation

The stat component has a feature where you can hover over it to display additional information. The hover triggers a transition on the component that causes it to move upwards by 10 pixels. While the transition itself is a smooth movement, it appears to cause the parent container to increase its height abruptly and that causes content after the stat block to jump in a disconcerting way.

statSizeJump
Screen capture credit: @GaryRidgway 

# Proposed solution
Apply a transition to affect the height of the stat container so that the movement to the new height appears smooth like the movement of the stat itself.

bspeare:
Resolves https://github.com/uiowa/uids/issues/947. 

# How to test

1. Test `stat_hover` in all of the Stat stories. 
2. Confirm new `stat_hover` control works. 

- Chrome transition will be very smooth
- Safari and Firefox will be better than it currently is but not as smooth as Chrome


GaryRidgway:
Pulling for review

GaryRidgway:
Made a tiny change to give us some flexibility. We should be able to use it in all browsers now!

e-marie-w:
Stat hover works and looks good, I tested in all three browsers mentioned. I see we have a story for Grid but it is not listed in the display property like default and horizontal are. I can try to get that working.

e-marie-w:
Wait default and horizontal can be applied to the grid, nevermind.

e-marie-w (approved):
Everything looks and works as expected

bspeare:
Thanks @e-marie-w!

PR #950, opened 2024-12-03
4.x - Banner

james917:
Related to https://github.com/uiowa/uids/issues/721. 

## How to test

- Code review
```
nvm use && yarn install && yarn storybook
```
- Review banner controls and confirm that they are all functional and control the property they are meant to control.
- Review banner variants (stories). Review prop/control defaults for each story and provide feedback if they don't make sense.
- Test banner control combinations to ensure options work together.

### Specific things to check:
- A new stories file was added for Media. Confirm that it is not displayed in the Storybook navigation.
- When there is only one button, the link text is rendered as a pseudo button and the banner title is linked.
    - Click to view the full page view of the banner and confirm that entire banner is clickable with the `click-container`. 
 - Confirm when there is more than one button that the links are directly applied to the buttons and not to the banner title.

PR #964, opened 2025-01-06
[Banner] Changes after more testing in uids_base

bspeare:
# How to test
- `yarn storybook`
- Compare http://localhost:6006/?path=/story/components-banner--background-image&args=mobile_content_below_image:!false  against https://uids.brand.uiowa.edu/latest/?path=/story/components-banner--background-image&args=mobile_content_below_image:!false.  Check out the banner on mobile.  Confirm the headline is white instead of black on mobile. 
- Try adjusting background combinations and controls and look for any other issues with color contrast. 

This pr sets the banner headline `<a>` to white by default and then is overridden in by all other background color and gradient variants.  In https://github.com/uiowa/uids/pull/950, we removed the `headline--negative` class which was used to set the headline color on dark backgrounds.

### DIFF
--- a/src/scss/components/banner.scss
+++ b/src/scss/components/banner.scss
@@ -843,6 +843,8 @@
     z-index: 1;
 
     a {
+      color: #fff;
+
       &:hover {
         text-decoration: none;
       }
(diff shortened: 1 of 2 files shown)


joewhitsitt (approved):
This looks good.

PR #966, opened 2025-01-13
Accordion - Support other heading sizes

joewhitsitt:
Because our collection block in uiowa supports parent/child heading nesting, accordions could have different heading sizes and these should be accounted for in the styles.

# Test

- Pull down and inspect an accordion element. Change the DOM for the accordion from h2 to h3 or greater. See that the styles don't break with the change. Aka, you don't want to see the following:

Screenshot 2025-01-13 at 9 38 43 AM

### DIFF
--- a/src/scss/components/accordion.scss
+++ b/src/scss/components/accordion.scss
@@ -40,7 +40,7 @@
     }
   }
 
-  .accordion__heading h2 {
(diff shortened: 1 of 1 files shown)


bspeare (approved):
Looks good!

PR #967, opened 2025-01-21
Accordion - move icon stuff into summary, consolidate rules

joewhitsitt:
https://github.com/uiowa/uiowa/pull/8405#issuecomment-2604972351

I don't have an easy way to test this in UIDS but by moving the icon styles under summary, it prevents it from changing icons elsewhere in the details element which contains the accordion body.

# To Test

Ensure the UIDS version is still working as expected. The icon rotates and is positioned correctly when the accordion item is expanded, collapsed.


bspeare (approved):
Looks good!

PR #970, opened 2025-01-23
Set banner to hidden

bspeare:


src/components/banner/banner.config.yml
```diff
@@ -36,7 +36,7 @@ context:
 variants:
   - name: default
     label: Large, content centered
-    hidden: false
+    hidden: true
```
pyrello, line 39: Do we need to set this on variants? If so, does it need to be set on the other variants?

bspeare: Good catch! Thanks!

pyrello (approved):
Reviewed, not tested.

PR #983, opened 2025-04-03
[UL/LI] Line-height adjustments

### LINKED ISSUE #990: Unordered/ordered lists within tables in accordions have a different font size than other cells
# Problem
If you have an unordered or an ordered list within a `<td>` element inside an accordion, it will be a different font size (`1.3rem`) than the text contained inside other `<td>` elements (unless you have `<p>` tags within each cell, but it appears the WYSIYWG strips most `<p>` elements out of `<td>` elements unless there are multiple paragraphs.

Image

# Steps to reproduce
1. In Layout Builder, create a Collection block, set it to "Accordion"
2. Add a `<table>` element with at least a row and two columns.
3. In one column, just put any single line of text
4. In the second column, add an ordered or unordered list
5. Note the discrepancy in font size between them (see screenshot above)

# Proposed solution
Investigate the use of the following CSS in [accordion.scss](https://github.com/uiowa/uids/blob/4.x/src/scss/components/accordion.scss#L84)

```
.accordion__content {
  ul li,
  ol li,
  p {
    font-size: 1.3rem;
  }
}
```
...and try to account for the fact that tables (and thus table cells) may be housed within an accordion ([even if the docs say this is discouraged](https://sitenow.uiowa.edu/news/2025/02/get-your-collections-order-guide-using-component-wisely#:~:text=Limit%20the%20use%20of%20tables%20in%20the%20collection%20component)-- it can still happen!) and may have content without any surrounding `<p>` tags. If it's necessary to keep content within accordions `1.3rem` in font size, then the above CSS could be adjusted to be:

```
.accordion__content {
  ul li,
  ol li,
  td,
  p {
    font-size: 1.3rem;
  }
}
```

bspeare:
Resolves https://github.com/uiowa/uids/issues/990. 

This pr contains a few line-height adjustments for our base list styles.  Brand requested that we tighten up the line-height.

### DIFF
--- a/src/scss/base/_lists.scss
+++ b/src/scss/base/_lists.scss
@@ -28,10 +28,6 @@ ol:not([type]).is-upper-roman {
   list-style-type: upper-roman;
 }
 
-ol li {
-  line-height: 1.7;
-}
-
 ul {
   list-style: disc outside;
   font-size: variables.$content-font-size;
@@ -40,14 +36,14 @@ ul {
   margin-left: 1.875rem;
 }
 
-ul li {
+ul li,
+ol li {
   line-height: 1.5;
   margin-bottom: 0.25rem;
 }
 
 ul ul {
   list-style-type: circle;
-  //@include margin($top: $xsm);
 }
 
 ul ul ul {
--- a/src/scss/base/_lists.scss
+++ b/src/scss/base/_lists.scss
@@ -39,7 +39,7 @@ ul {
 ul li,
 ol li {
   line-height: 1.5;
-  margin-bottom: 0.25rem;
+  margin-bottom: 0.5rem;
 }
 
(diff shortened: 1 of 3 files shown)


joewhitsitt:
pulling to test

joewhitsitt:
>Brand requested that we tighten up the line-height.

`¯\_(ツ)_/¯` to the naked eye, it looks more "loose" but i see from the code changes it is tightened.

Screenshot 2025-06-12 at 10 22 07 AM

joewhitsitt:
>Review and approve https://github.com/uiowa/uids/pull/983.

Approved.

joewhitsitt (approved):

bspeare:
Thanks @joewhitsitt!

PR #986, opened 2025-04-15
[Bug] Banner play button

bspeare:
Related to https://github.com/uiowa/uiowa/issues/8630. 

# How to test
1. Code review
2. Confirm changes that videos still autoplay on http://localhost:6006/?path=/story/components-banner--grid-video and if used in a browser where videos are set to not autoplay - they will not autoplay.


cory-skeers (approved):
Reviewed and tested--autoplay respects the browser settings, and the button matches the action.

PR #993, opened 2025-07-11
[UIDS] Increase spacing on nested list items 

### LINKED ISSUE #991: Increase spacing on nested `<ol>` and `<ul>` list items
# Issue

We need to increase spacing on nested `<ol>` and `<ul>` list items ~and we need to reduce the bottom margin for `<p>` tags within `<ol>` and `<ul>` list items~. 

## Example possible solution

```
ol ol, ol ul, ul ol, ul ul {
    margin-block-start: .75rem;
    margin-block-end: .75rem;
}

```

**Before:**

Image

**After:**

Image

**Before:**

Image

**After:**

Image

e-marie-w:
Closes #991 

## Testing Instructions
`yarn storybook`
In the blockquote's content control (or a similar place where you can add html), add some nested list items, such as: 
```
<ol><li>This is the first item in the original ordered list.</li><ol><li>This is the first item in the nested ordered list.</li><li>Second item in nested ordered list.</li></ol><li>Second item in original ordered list.</li></ol>
```
You can try different combinations of `<ol>` and `<ul>` if you want. Spacing difference should be subtle but there

### DIFF
--- a/src/scss/base/_lists.scss
+++ b/src/scss/base/_lists.scss
@@ -52,9 +52,9 @@ ul ul ul {
 
 ol ol, ol ul, ul ol, ul ul {
   margin-top: 0;
-  margin-block-start: 0;
+  margin-block-start: .75rem;
   margin-bottom: 0;
-  margin-block-end: 0;
+  margin-block-end: .75rem;
 }
 
 .element--list-none {


joewhitsitt:
pulling to test

joewhitsitt (approved):
Looks good to me. I think it is scoped correctly and if it was to leak out, it wouldn't be much of an issue.

Issue #997, opened 2025-07-17
Create application form layout in UIDS


bspeare:
- [ ] Recreate profiles directory profiles.uiowa.edu layout (screenshot below) in UIDS Storybook in order to provide feedback on https://ais-components.its.uiowa.edu/.   The form styles from the previous slice should provide a base for the form elements.
- [x] Create Menu component with horizontal and vertical variants.  The styles from uids_base will need to be brought over to UIDS in order to build this component: https://github.com/uiowa/uiowa/blob/main/docroot/themes/custom/uids_base/scss/components/menus/main-menu.scss

<img width="3178" height="2884" alt="Image" src="https://github.com/user-attachments/assets/ae90fc31-31d8-4f13-91eb-5264754cb4ba" />

briand44:
Grooming conversation... Shouldn't need to style WYSIWYG. Can just have a text area there. Follow the tasks listed in the IS. We may need to account for the icon (left) on the button component. Rule for adding spacing when adjacent icon. Brand bar, top nav, sidebar nav. Recreate best we can without getting into JS.

bspeare:
I wonder as part of this if we can get the sidebar away from the edge so that it is placed in a container and aligned under the brand bar.  Something like this:

<img width="1537" height="1106" alt="Image" src="https://github.com/user-attachments/assets/9a0ccab6-dead-47f1-b530-ad3923fc5ff3" />

joewhitsitt:
Could it follow our same layout logic and just fill the sidebar with the grey instead of bleed out to the layout margins? Padding the select and radio inputs of course.

<img width="1371" height="815" alt="Image" src="https://github.com/user-attachments/assets/a6a9a82f-e5d8-48d2-9e79-bec834b452d6" />

bspeare:
@joewhitsitt yes, I think that is a better idea.

bspeare:
<img width="1485" height="1078" alt="Image" src="https://github.com/user-attachments/assets/df39a09d-342e-40e4-bc1a-6fc168866af0" />

Denny also mentioned today that it would be useful to have a mockup of this page with exposed form controls and table.  Sidebar form controls on https://classrooms.uiowa.edu/spaces/find-university-classroom and top controls on https://iowasummerwritingfestival.uiowa.edu/events would be a good place to start.

joewhitsitt:
Grooming convo: Basic scaffolding can be done but ultimately needs to wait for the other slice as it requires the styles from that to finish.

[...]

PR #998, opened 2025-07-18
Example info request form

### LINKED ISSUE #995: Create form example page in UIDS
AIS is looking to create a 1.0 release of https://ais-components.its.uiowa.edu/ and would like feedback on their components. 

# Possible Slices

- [ ] Recreate https://www.maui.uiowa.edu/maui/pub/admissions/webinquiry/undergraduate.page (screenshot below) in UIDS Storybook combining https://github.com/uiowa/uiowa/blob/main/docroot/themes/custom/uids_base/scss/components/form/forms.scss and https://github.com/uiowa/uiowa/blob/main/docroot/themes/custom/uids_base/scss/webforms/webforms.scss to create form styles in UIDS. 

Image

GaryRidgway:
[I have a branch here that I am using as a sort of checklist to go through and move styles over.](https://github.com/uiowa/uiowa/pull/9048)


bspeare:
Remaining work here is to make the focus styles stronger and get review from @pyrello.

pyrello:
> Remaining work here is to make the focus styles stronger and get review from @pyrello.

I'm going to review this today.

pyrello:
Some initial thoughts:
- Because the form being created in this PR is a mockup of an existing form and its purpose is to help guide providing styles for that particular form, I don't think it should be built in a component called Form.vue. If we are going to use a component, maybe we need a component with a name specific to that form.
- I'm not sure that creating a Vue component is warranted here. The implication of a Vue component in the context of a library like this is that someone could consume it. But there is no use case for consuming a demo form.
- If we were going to use a form component to drive this, the form component itself would want to be agnostic to the specific form and be able to render different forms depending on the input being passed to it. There might already exist a Vue plugin that does something like this so it would be worth looking to see so we're not reinventing the wheel.
- I'm open to the above being part of this slice (or at least some experimenting to see how difficult it is), but if we don't end up using a form component here, I would suggest we just hard code the form markup into a story file.
- Similar to above, I don't think we should be using a Form.stories.js file for this specific demo form. While I think this could morph into being a form that is used to demo our form styles, this hasn't been decided yet and I don't want this form to become the de facto form component without some thinking about whether that is what we want or not. To that end, I'm wondering if we need a `demo` or `showcase` namespace for providing these types of mockups and examples.

pyrello:
@GaryRidgway and I discussed my [comments above](https://github.com/uiowa/uids/pull/998#issuecomment-3128247698). Based on that conversation, here is my understanding of next steps:
- Create a `showcase` folder inside the components folder. The purpose of this is to have a place to put mockups or other examples that combine multiple components in more complex configurations.
- Build a `.stories.js` file for the example form that is part of this work. The first iteration of this is that all the form markup just lives in this file. If we find value in it later, we can swap out the markup with imported components.
- The `Form.vue` file can be removed for now or its template could be replaced with something like the following that just allows it to be a convenient way to wrap form elements in a form tag:
```html
<form>
  <slot></slot>
</form>
```
- The mockup being created in #997 can be implemented in a similar way as above, but potentially with some of the parts of the template in the `.stories.js` file being replaced with Vue components if the work progresses that far.

pyrello (approved):
This looks good and provides a good starting point for adding form styles as well as a nice looking example form.

PR #1012, opened 2025-10-17
update element-invisible utility

joewhitsitt:


joewhitsitt:
If we move forward with this, keep in mind the uids_base tweaks needed to prevent regressions with the addition of clip-path where it is not expected.

https://github.com/uiowa/uiowa/pull/9280/commits/49aea168c4e9566f72eb007137f50049d2750b3c

PR #1013, opened 2025-10-22
adjust stat and table as it relates to all caps text

joewhitsitt:
Relates to https://github.com/uiowa/uiowa/pull/9304

See https://github.com/uiowa/uiowa/pull/9304 for testing instructions.

~~`capitalize` is i think the only css option here. so if `BIG AND BOLD` is the value, it will respect it and leave it as all caps. If the value starts with a lower case letter `living On Campus - Example Table` it will change it to `Living On Campus - Example Table`~~ though this is true, not the way to move forward with this.

### DIFF
--- a/src/components/stat/Stat.vue
+++ b/src/components/stat/Stat.vue
@@ -66,7 +66,7 @@ const classes = computed(() => {
       </h2>
     </div>
 
-    <p v-if="stat_summary" class="stat__description" v-html="stat_summary"></p>
+    <span v-if="stat_summary" class="stat__description" v-html="stat_summary"></span>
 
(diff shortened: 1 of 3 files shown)


joewhitsitt:
Read online that the css hack seems to be 

```css
.sentence-case {
  text-transform: lowercase;
}
.sentence-case::first-letter {
  text-transform: uppercase;
}
```

**Update**: 

>Personally, I don't think we should be forcing people into sentence case. Its fine to remove uppercasing, but I think there will be unintended consequences for forcing sentence case. - Sean

cory-skeers (approved):
Reviewed, and tested as part of https://github.com/uiowa/uiowa/pull/9304 👍

PR #1018, opened 2025-11-21
[Tables] Switched to max-height to pass Siteimprove scan

bspeare:
Related to https://github.com/uiowa/uiowa/issues/9366. 

- Undo the file change for `max-height` and run the Siteimprove plugin on http://localhost:6006/iframe.html?globals=&args=&id=components-table--default&viewMode=story.  Confirm you see the "Scrollable element is not keyboard accessible" issue. 
- Redo the file change and test the page again and the error should be gone. 
- Review https://github.com/uiowa/uiowa/pull/9398. 
- A11y fix taken from `max-height` example from https://codepen.io/aardrian/pen/VwYyVJY from https://adrianroselli.com/2020/01/fixed-table-headers.html.

### DIFF
--- a/src/components/table/Table.vue
+++ b/src/components/table/Table.vue
@@ -41,7 +41,7 @@ const tableClasses = computed(() => {
 </script>
 
 <template>
-  <div class="table-responsive" :class="{ 'table--sticky': props.sticky }">
+  <div class="table-responsive" :class="{ 'table-sticky': props.sticky }">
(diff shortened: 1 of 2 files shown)


cory-skeers (approved):
Nice update, and appreciate the inclusion of background info in finding the solution!

bspeare:
Thanks @cory-skeers!

PR #1023, opened 2025-12-10
White accordion option

GaryRidgway:
# How to test

- Code review
- Test locally or at https://uids.brand.uiowa.edu/branches/gold-accordion/?path=/story/components-accordion--multi-select&args=background:white&globals=backgrounds.value:!hex(f3f3f3). 

## Review new white option
- Test with background colors using the highlighted icon below. 
Screenshot 2026-01-12 at 3 00 47 PM
- Test focus/active styles and compare against https://uids.brand.uiowa.edu/?path=/story/components-accordion--default. 

## Review updated gray option
- Accordion gray now uses the same background color gray.
- A border has been added to create distinction between background gray.
- Test focus/active styles compare against https://uids.brand.uiowa.edu/?path=/story/components-accordion--default. 

## Review new arrow placement
- Does the spacing look correct?
- Any concerns on https://github.com/uiowa/uids/pull/1023#issuecomment-3716073790?
- Note that the active white circle background has been removed.
- Consider wider change impact such as the accordions on https://uipress.uiowa.edu/books/brightening-glance will have their arrow placement changed.
## Grey to gray
- Updated usage of `grey` to `gray` to reflect American standard usage used by brand manual.


joewhitsitt:
~~pulling to test~~

wait, realized that this might not be ready for review yet.

bspeare:
Sorry @joewhitsitt, this is not set as draft so we can share the dev url with brand team.  I should have posted that here.

pyrello:
Set to draft until we're clear on next steps.

bspeare:
A few findings on icon placement:

https://carbondesignsystem.com/components/accordion/accessibility/

> Carbon chevrons are right-aligned by default, but left-aligned chevrons are more accessible for users with low vision, as the expanded/collapsed indicator is closer to the accordion title.

https://www.srishti.design/accordions#comp-m0f3au2g
Screenshot 2026-01-06 at 1 35 40 PM

https://www.viget.com/articles/testing-accordion-menu-designs-iconography

> Analyzing the results, the most convincing data involved icon location. Presented with an icon to the right of the menu item, many participants preferred to click the icon rather than the text label (Figure 2). In addition, in every case, designs with icons to the right resulted in slower task times. 

https://practical-accessibility.today/chapters/accordion/
- Uses left icon position

https://adrianroselli.com/2023/08/progressively-enhanced-html-accordion.html
- Uses left icon position

bspeare:
Examples using arrows and icons:

https://www.shadcn.io/patterns/accordion-subtitle-2

Screenshot 2026-01-06 at 1 48 15 PM

The left arrow does open the use of adding a badge to the right of the accordion for status and category indications.

bspeare:
Updated branch with left icon: https://uids.brand.uiowa.edu/branches/gold-accordion/?path=/story/components-accordion--multi-select&args=background:white&globals=backgrounds.value:!hex(f3f3f3).

pyrello:
Pulling to test.

src/components/accordion/Accordion.stories.js
```diff
@@ -4,6 +4,20 @@ export default {
   title: 'Components/Accordion',
   component: Accordion,
   argTypes: {
+    background: {
+      name: 'Background',
```
pyrello: Background doesn't seem like the right term for this since it doesn't apply to the entire component. Wondering about "Handle color" or something like that instead.

src/components/accordion/Accordion.vue
```diff
@@ -57,12 +64,12 @@ onMounted(() => {
...
       >
         <h2>
-          {{ item.title }}
           <i aria-hidden="true" class="fas fa-chevron-up" role="presentation"></i>
+          {{ item.title }}
         </h2>
       </summary>
       <div
-        class="accordion__content"
+        class="bg--white accordion__content"
```
pyrello: The implication of using `bg--white` to apply the background color of this element is that folks looking at this might conclude that they could just replace that class with another `bg--*` class to change the color.

src/components/accordion/Accordion.vue
```diff
@@ -2,13 +2,19 @@
...
+import '../../scss/components/_background.scss'
 import {ref, computed, defineProps, onMounted} from 'vue';
 
 const props = defineProps({
   multiselectable: {
     type: Boolean,
     default: false,
   },
+  background: {
+    type: String,
```
pyrello, line 14: How confident are we feeling that we will not add any more colors here? Any chance of a black accordion handle? If we think there is a chance of more colors being added, I think this is okay. If we feel like there will not be more colors added, then I might go with a Boolean here instead, which would be consistent with how you are adding the class below.

src/components/accordion/Accordion.stories.js
bspeare: That is a good point.  Tab color, label color, button color, or handle color?

src/components/accordion/Accordion.vue
bspeare: Not entirely confident about no more colors.  I think this is just where Alan started.

src/components/accordion/Accordion.vue
bspeare: Good point.  Let's directly set the white background without using that class.

src/components/accordion/Accordion.stories.js
pyrello: Label color is the one I vibe with the most, I think.

src/scss/components/accordion.scss
```diff
@@ -3,69 +3,82 @@
...
+    --accordion-color: #{variables.$white};
+
+    summary {
+      &:hover,
+      &:focus {
+        --border-color: var(--uids-gray-cool);
+        background-color: var(--uids-light);
+      }
+    }
+  }
```
pyrello: I don't think we benefit from this being nested here as opposed to just have it as it's own block later. I prefer the pattern of having modifier classes being later in the CSS file after the base definition. This makes the file easier to read because you can see what the default styling is first then see how the modifier changes it.

src/scss/components/accordion.scss
```diff
@@ -3,69 +3,82 @@
...
   &:not(:first-child) {
     margin-top: 1rem;
   }
 }
 
 .accordion {
   --trans-time: 0.2s;
-  $accordion-grey: #edeceb;
-  --accordion-grey: #{$accordion-grey};
+  --accordion-color: var(--uids-light);
```
pyrello: Let's adjust this variable name to correspond to "label color" verbiage discussed in another comment.

src/scss/components/accordion.scss
```diff
@@ -106,4 +88,27 @@
...
+
+// === Accordion modifiers === //
+.accordion--white {
+  --accordion-label-color: #{variables.$white};
+
+  summary {
+    &:hover,
+    &:focus {
+      --border-color: var(--uids-gray-cool);
+      background-color: var(--uids-light);
```
pyrello: You should be able to change this by setting the variable instead of the `background-color` since it is set to the variable in the base definition. Similar to how you changing `--border-color`.

src/scss/components/accordion.scss
```diff
@@ -2,79 +2,74 @@
...
     list-style: none;
-    width: 100%;
-    background-color: var(--accordion-grey);
+    background-color: var(--accordion-label-color);
     display: flex;
     cursor: pointer;
 
     &:hover, &:focus {
-      background-color: color.scale($accordion-grey, variables.$lightness: -4%);
+      background-color: variables.$lighter-gray;
```
pyrello: This is another place where `background-color` can be changed by setting `--accordion-label-color`. Alternately, we could set another variable called `--accordion-label-color-hover` here and remove the need to override `:hover` and `:focus` in the `accordion--white` section because you could just set the value of `--accordion-label-color-hover` as part of the `summary` definition there.

pyrello:
Some thoughts after more hands-on testing. The contrast between the white and the gray hover state color is pretty stark. To me it feels a little bit jarring. I'm wondering if we have any other less dramatic color we could use as a hover color? Similarly, the border/outline color seems too vivid. Since we don't use a lot of borders that dramatic, it makes the white accordion look a bit out of step with the rest of our branding choices. Wondering if we could use a lighter color to make the effect less?

pyrello:
I updated `Accordion.stories.js` to adopt syntax for shared template and best practice for individual story definition.

bspeare:
@pyrello I updated the white focus state.  I do like the more dramatic outline so that you can distinguish it against other accordions.  Do you want to play around with that and see if you find something that feels better?  Seems like we need another gray that is not as dark, I'm just not sure if we want to introduce another gray.

pyrello (approved):
I think this is a solid improvement in terms of adding a new option but also keeping the relevant files up to date with our current best practices.

PR #1027, opened 2025-12-18
[Headline] Added overflow property for headline underline

bspeare:


joewhitsitt (approved):
Approving based on https://github.com/uiowa/uiowa/pull/9458's instructions. Does the trick.

Issue #1021, opened 2025-12-05
Finish defining form styles


pyrello:
- [x] Fix alignment on the Toggle control.
- [x] Review all form elements in Storybook and test all combinations of SB controls on each element. Make note of any cases that are not currently handled or styles need improvement.
- [x] When we think we are at a good point, let Denny know how he can pull a local version to test and give feedback. This should happen at a point in the process where there is enough time to make changes based on that feedback.
- [x] Should the `required` attribute be included when appropriate? If so, on which controls?
- [x] Adjust button small and medium sizes to use the --form-input-height so the sizing matches the fields.  Maybe also need to tweak mobile behavior of buttons when used in a form.
- [x] The "Compact" option gets a a little too squished.  Maybe the default is compact and we flip this to "Large gap" or something to pair with the "Large" modifier?
- [x] Create a follow-up issue in `uiowa` for integrating new form styles. 
- [x] Review all CSS class names. Make updates to ensure consistency where appropriate. Example: should we use `form--required` instead of `form-required`?
- [x] Examples for side-by-side for checkboxes and radios: https://github.com/uiowa/uiowa/blob/main/docroot/themes/custom/uids_base/scss/components/form/forms.scss#L219 and https://github.com/uiowa/uiowa/pull/9461 for reference. 
- [x] Should icons have an explicit footprint in our form styles and components (as opposed to just being something that can be inserted anywhere text would be appropriate)?
  - Decision: Icons can currently be handled by including them as text in the label of a form field. Need more info and examples of how that falls short.
- [x] Error, focus, and validation state styling for form controls.
  - Decision: Other validation states than error aren't currently used by us. This can easily be layered on later if there is value.

bspeare:
# Grooming Conversation

**Quadrant:** 
- Complex (some things are yet tbd possibly)

**Discussion:**

- A task to do our regular process of component definition for form styles. Some of the work has already been done, but we are going to be intentional with the process here to cover as may bases as possible or reasonable. 
- Is this something where we want to solicit feedback from AIS? 
  - Look for feedback and make changes in future slices. 

**Proposed solution or definition of done:**

- A set of form styles that are ready to be shared with campus partners.

pyrello:
Feedback from testing/review of #1016: https://github.com/uiowa/uids/pull/1016#issuecomment-3647327009

pyrello:
[...]

PR #1029, opened 2026-01-05
Form control stories

### LINKED ISSUE #1020: Create toggle/switch checkbox example for the kitchen sink page
Add a toggle/switch checkbox example for the kitchen sink page

- Markup at https://github.com/uiowa/uids/pull/1016/commits/4b532ec95aba168bd2896386a52eeeadb7ede786
- CSS at https://github.com/uiowa/uids/blob/823f05c5c5626ec1aab9a5b17571ff237a756eb9/src/scss/components/form.scss#L334

### LINKED ISSUE #1022: Build out remaining storybook stories and controls for forms
- Maybe create separate kitchen sink pages?  Using the "Inline" and "Flex" control with text, email, phone, search, and select is a good example if using that property.
- It would be great if showSidebar control on http://localhost:6006/?path=/story/showcase-application--directory would move the form into the sidebar to showcase the container queries.
- Implement new classes on http://localhost:6006/?path=/story/showcase-application--form?  Also add a save button to this form?
- Same with http://localhost:6006/?path=/story/showcase-forms--information-request-form, rebuild this with new tools from the kitchen sink and see how far we can make it without the flexbox showcase styles? Or maybe building out a form layout will tell us if we need any other helper classes.

pyrello:
Resolves #1022 
Resolves #1020 

## To test
- Code review.
- Review all form elements in storybook.
- Test out storybook controls for each element. Each SB control should result in a change to the markup/classes on the form element even if you can't see an obvious visual change.
- Note anything that you think is missing.


joewhitsitt:
pulling to test

joewhitsitt:
should required state add the attribute to the input in addition to the label `form-required` class? https://www.w3schools.com/tags/att_input_required.asp

i am seeing it on some elements but not on all.

pyrello:
> should required state add the attribute to the input in addition to the label `form-required` class? https://www.w3schools.com/tags/att_input_required.asp

Good question! I think this is something we will punt to #1021

joewhitsitt:
> > should required state add the attribute to the input in addition to the label `form-required` class? https://www.w3schools.com/tags/att_input_required.asp
> 
> Good question! I think this is something we will punt to #1021

I want to clarify I am not talking about the styles. When the toggle is triggered it is not adding the attribute to the input for checkbox, text, file, etc. It is being added to radio and some others, which what I thought was being tested and included in this PR?

Also for some of the controls, I am not seeing a markup change for inline. http://localhost:6006/?path=/story/elements-form--checkbox&args=required:!true;inline:!true
http://localhost:6006/?path=/story/elements-form--select&args=inline:!true

joewhitsitt:
>Each SB control should result in a change to the markup/classes on the form element even if you can't see an obvious visual change.

For testing I am toggling each thing on each control and seeing if there is a change to the markup. If I am going about this wrong, please let me know.

I am also checking to see if I think a state or modifier I see on one control should be represented on another.

joewhitsitt:
http://localhost:6006/?path=/story/elements-form--file and the button/reset/submit doesn't have an inline modifier toggle. should they?

pyrello:
> > Each SB control should result in a change to the markup/classes on the form element even if you can't see an obvious visual change.
> 
> For testing I am toggling each thing on each control and seeing if there is a change to the markup. If I am going about this wrong, please let me know.
> 
> I am also checking to see if I think a state or modifier I see on one control should be represented on another.

This sounds like a great way to go about it!

> http://localhost:6006/?path=/story/elements-form--file and the button/reset/submit doesn't have an inline modifier toggle. should they?

The buttons currently don't default to being full width like the rest of the form controls, so in that context, I didn't include the inline option for them, as it doesn't do anything to them.

pyrello:
> > > should required state add the attribute to the input in addition to the label `form-required` class? https://www.w3schools.com/tags/att_input_required.asp
> > 
> > 
> > Good question! I think this is something we will punt to #1021
> 
> I want to clarify I am not talking about the styles. When the toggle is triggered it is not adding the attribute to the input for checkbox, text, file, etc. It is being added to radio and some others, which what I thought was being tested and included in this PR?

I had not previously come across the `required` attribute in my due diligence. If it exists on some elements currently and not others that is an artifact of copy/paste, which probably originated in SiteNow. I would like to spend some time understanding the relevance of the `required` attribute to determine whether it is important to include and in which cases. As such, I think it should be handled in the follow-up issue as part of the due diligence necessary for that task.

joewhitsitt:
No markup change for error on http://localhost:6006/?path=/story/elements-form--select-multiple&args=error:!true

pyrello:
> No markup change for error on http://localhost:6006/?path=/story/elements-form--select-multiple&args=error:!true

> Also for some of the controls, I am not seeing a markup change for inline. http://localhost:6006/?path=/story/elements-form--checkbox&args=required:!true;inline:!true
> http://localhost:6006/?path=/story/elements-form--select&args=inline:!true

These should both be fixed now.

joewhitsitt:
- Thanks for fixing the checkbox inline, but there is also http://localhost:6006/?path=/story/elements-form--select&args=inline:!true (had updated the comment after initially posting).
- Thanks for the error control on http://localhost:6006/?path=/story/elements-form--select-multiple&args=error:!true
- You explained the button/reset/submit, but http://localhost:6006/?path=/story/elements-form--file doesn't have the inline control.
- I am fine with the state of the required attribute only being on some elements for now.
- As for the list of form controls themselves, there is nothing noticeably absent to me.

pyrello:
> * Thanks for fixing the checkbox inline, but there is also http://localhost:6006/?path=/story/elements-form--select&args=inline:!true (had updated the comment after initially posting).
> 
> * Thanks for the error control on http://localhost:6006/?path=/story/elements-form--select-multiple&args=error:!true
> 
> * You explained the button/reset/submit, but http://localhost:6006/?path=/story/elements-form--file doesn't have the inline control.
> 
> * I am fine with the state of the required attribute only being on some elements for now.
> 
> * As for the list of form controls themselves, there is nothing noticeably absent to me.

The additional instances where inline have been remedied.

joewhitsitt (approved):
Thanks for the additional tweaks

PR #1038, opened 2026-03-25
Scope video script to banner only

pyrello:
# Motivation

The video is not scoped to the banner component, but it was never really intended to be applied to other videos. This PR scopes the script so that only banner videos are targeted.

# How to test

```
nvm use && yarn storybook
```
- http://localhost:6006/?path=/story/components-banner--background-video
- Confirm that video still functions as expected.
- To check that this fixes the console error on the homepage site, you can manually make the same change to the docroot/themes/custom/uids_base/uids/assets/js/video.js file in the SiteNow repo and check it locally.

### DIFF
--- a/src/assets/js/video.js
+++ b/src/assets/js/video.js
@@ -6,7 +6,7 @@
   function Video(element, index) {
     if (element) {
       this.element = element;
-      this.video = this.element.querySelector('video');
+      this.video = this.element.querySelector('.banner video');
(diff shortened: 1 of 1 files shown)


GaryRidgway:
Pulling to test

GaryRidgway (approved):
Did as the testing instructions described and found this fix worked as expected!

PR #1088, opened 2026-08-27
5.x - Add design tokens (primitives + semantic)

### LINKED ISSUE #1024: 5.x - Add primitive and semantic tokens
# Motivation

Tokens represent one of the most portable units of transferable branding information. 4.x has never previously had a formal implementation, just informal CSS variable tokens. By providing tokens in 4.x, we can unlock the potential to have these feed into other systems and be used by AI agents.

# Proposed solution

- Retrofit (backfill) implementation of existing, relevant tokens. Excludes component-specific tokens or one off hard-coded values.
- Forward-looking naming structure. Tokens will be named in accordance with how we expect to name things going forward.
- Primitive and semantic token layers included.
- Swap out existing token references.

## Not in scope
- Component tokens.

pyrello:
## Summary

Introduces UIDS's first authoritative global token layer for color, typography, spacing, and layout.

- Defines DTCG 2025.10 token source in `src/tokens/`, organized as primitives and semantic roles.
- Generates CSS custom properties during the build with Style Dictionary.
- Adds semantic color roles, typography styles, and container widths, backed by global color, typography, and spacing primitives.
- Adds token tests and Storybook reference pages.
- Adds release packaging: publishing a GitHub release builds and attaches `uids-<version>.tgz`, which consumers can reference from `package.json`.
- Requires Node 22 or later.

## How to test

1. Install dependencies and build:
```bash
nvm use
yarn install --immutable
yarn build
```
- Confirm that the build completes successfully.
- Confirm that the generated token Sass partial is created at `src/scss/abstracts/_tokens-generated.scss`.
- Confirm that `dist/tokens.css` is generated.
- Confirm that `dist/uids.css` and `dist/uids-core.css` contain the generated custom properties.
- Run the unit tests and confirm they pass:
```bash
yarn test:unit --run
```

2. Start Storybook:

```bash
yarn storybook
```

Review the [Colors tokens page](http://localhost:6006/?path=/story/tokens--colors), [Typography](http://localhost:6006/?path=/story/tokens--typography), [Space and layout](http://localhost:6006/?path=/story/tokens--space-and-layout), and the [web-page showcase](http://localhost:6006/?path=/story/showcase-web-page--web-page). The showcase is particularly useful for checking nested background and foreground combinations.

To verify token watching, temporarily change a token value while Storybook is running and confirm that the affected example updates without restarting Storybook. Revert the temporary change afterward.

3. Confirm the package contains the intended deliverables:

```bash
npm pack --ignore-scripts --dry-run
```

The file list should include `dist/uids.css`, `dist/uids-core.css`, `dist/tokens.css`, a component stylesheet such as `dist/components/button.css`, `src/scss/abstracts/_tokens-generated.scss`, token JSON such as `src/tokens/primitives/colors.json`, and files from `src/assets/`.

4. Optional cleanup

- If you need to run through any of these checks again you can run the following to remove the generated files:
```bash
rm -rf dist src/scss/abstracts/_tokens-generated.scss node_modules/.cache/storybook
```

## Expected rendered changes

`CHANGELOG.md` records every intentional rendered-value change. The main changes are:

| Group | What changes |
| --- | --- |
| Spacing scale | Four Sass variables move to a 4px grid, affecting padding and margin across components |
| Heading sizes | Heading maximums change and `h6` becomes fluid; minimums remain unchanged |
| Serif headings | No longer carry an independent `font-size` at `h2` through `h6` |
| Status colors | All eight status and status-surface colors change; warning no longer matches danger |
| Grays | Three neutral steps move to the Foundations palette values |

## Release behavior

The release workflow requires a GitHub release tag matching `package.json` (for example, `v5.0.0-alpha.0` for package version `5.0.0-alpha.0`). It builds the tagged source and uploads `uids-<version>.tgz` to that release. See the repository README for the release procedure.


bspeare:
There are a few things that would be nice to have in here that are currently in uids_base. 

**Layout & Spacing:** 
- It would be great to have this represented in UIDS so that we can have it in the tokens, UIDS, and Figma for testing out changes the design team would like to make to the layout and spacing system.  
  - Example in (https://github.com/uiowa/uids/pull/1087) http://localhost:6006/?path=/story/foundation-layout--all-layout.         
  - Also, testing equal height components in a layout is also useful: https://github.com/uiowa/uids/blob/figma_tokens_uids/src/scss/layout/_grid.scss#L111

**Tabs, Pager, Breadcrumbs, Page title:**
- We will need to test typography, color and spacing changes and having these in the tokens, UIDS, and Figma would be ideal for building out landing pages (http://localhost:6006/?path=/story/showcase-biochemistry-landing-page--default).  
- Examples in (https://github.com/uiowa/uids/pull/1087) 
  - http://localhost:6006/?path=/docs/components-tabs--docs. 
  - http://localhost:6006/?path=/docs/components-pager--docs
  - http://localhost:6006/?path=/docs/components-page-title--docs 
  - http://localhost:6006/?path=/docs/components-breadcrumbs--docs

pyrello:
**Keep**

| Token | Value | 4.x name |
|---|---|---|
| `gray.100` | `#F3F3F3` | `$light` / `--uids-light` |
| `gray.200` | `#CACACA` | `$gray-light` / `--uids-gray-light` |
| `gray.300` | `#9A9A9A` | `--uids-gray-mid` |
| `gray.400` | `#737373` | `$gray` / `--uids-gray` |
| `gray.500` | `#63666A` | `$brand-cool-gray` / `--uids-gray-cool` |

**Drop**

- `gray.50` `#FAFAFB` (form.scss)
- `gray.115` `#EEEEEE` (tables.scss)
- `gray.150` `#E6E5E5` (brand-bar.scss, menu.scss)
- `gray.155` `#E6E6E6` (duplicates `$lighter-gray`)
- `gray.160` `#E1E1E1` (form.scss)
- `gray.165` `#DDDDDD` (tables.scss)
- `gray.175` `#CCCCCC` (brand-footer.scss, stat.scss)
- `gray.310` `#999999` (brand-footer.scss)
- `gray.350` `#8A8A8A` (form.scss)
- `gray.365` `#7C7C7C` (tables.scss)
- `gray.450` `#666666` (stat.scss)
- `gray.600` `#5D5656` (brand-footer.scss)
- `gray.700` `#484741` (form.scss)
- `gray.750` `#333333` (tables.scss)

**Why:** each is used by a single component, so it belongs in `tokens/components/*.json`, not the primitive tier. All 14 are referenced by nothing today (no semantic token, no authored SCSS), so removing them changes no output.

GaryRidgway:
I wanted to provide some examples of what I was thinking for using HSL or OKLCH instead of HEX for defining colors. I think it makes more sense with semantic or primitive tokens to use these. I think the values that we have been appending can almost directly map to the values in HSL or OKLCH. 

I've tried both OKLCH and HSL out, but I think HSL works enough. OKLCH is a bit more complicated to understand, but gives better chroma. I think the simplicity makes sense here.

HSL is representative of Hue Saturation and Lightness. You can edit any one of these independently.

For our uses, I would take the hue as a human read word, like "gold" and then use numbers to represent the other values. Here is what Iowa gold might look like if we have it in different shades.

Iowa gold:
- HEX: `#ffcd00`
- HSL: `hsl(48.24, 100%, 50%)`

Represented in tokenized form, we could remove the hue and use the word gold:

- `gold.100.50` : `hsl(48.24, 100%, 50%)`
- `gold.80.50` : `hsl(48.24, 80%, 50%)`
- `gold.60.50` : `hsl(48.24, 60%, 50%)`
- `gold.40.50` : `hsl(48.24, 40%, 50%)`

With greys, we could remove the saturation variable, as it is irrelevant:

- `neutral.100` : `hsl(0, 0%, 100%)`
- `neutral.80` : `hsl(0, 0%, 80%)`
- `neutral.60` : `hsl(0, 0%, 60%)`
- `neutral.40` : `hsl(0, 0%, 40%)`
- `neutral.20` : `hsl(0, 0%, 20%)`
- `neutral.0` : `hsl(0, 0%, 0%)`

src/components/showcase/BackgroundContexts.stories.js
```diff
@@ -0,0 +1,103 @@
...
+    </table>
+    <table class="table--gray-borders">
+      <thead><tr><th>Gray borders</th><th>Header</th></tr></thead>
+      <tbody><tr><td>Cell</td><td>Cell</td></tr></tbody>
+    </table>
+  `,
+};
+
+export default {
+  title: 'Showcase/Background contexts',
```
pyrello: The `Background Contexts` naming seems too specific to the usage we created this for. It probably has broader usages.

Also, it looks a little haphazardly produced and not very professional.

src/tokens/semantic/typography.json
```diff
@@ -0,0 +1,107 @@
+{
```
pyrello, line 1: Why is the file named `typography-usage` instead of `typography`?

pyrello:
Follow-up review summary: the primitive/semantic direction is sound, but I am tightening the token contract before approval. The responsive-heading rule should be explicit token metadata rather than a bare generator boolean; token groups should declare and validate their types; and semantic backgrounds should use one design-facing background hierarchy instead of CSS-flavored abbreviations. Component-specific decisions remain deliberately outside this initial global layer.

bspeare:
A few questions/observations:

Style dictionary
- What's scripts/style-dictionary.test.js covering? Note or comment at the top would help me. 
- style-dictionary.config.js could use a couple of comments to help me understand what's going on there at a glance. 
- Maybe just need more info in the Readme about it?

On extensions
- Only extension I see is edu.uiowa.fluid in semantic/typography.json. Is that the only one for now? Worth a line in the readme about how/why we are using that?

On the format docs
- README links the DTCG spec but doesn't say which bits we actually use. Could we get a short section on that? Or an example token file with comments above each one?

Fractal
- Fractal's still in README.md, lines 114 and 123 :) Pre-existing, not from this PR.

bspeare:
Do we need to be concerned about this example? https://github.com/uiowa/uids/blob/tokens_4x_primitives_semantic/README.md?plain=1#L108. I tested it and `--brand-primary` for the button CSS is not in `tokens.css`, do we need to specify that `uids.css` or `uids-core.css` needs to be added for that button example?

GaryRidgway:
I was looking at the representations of the primitive tokens, and I had a few questions:

- What is the reason for two primitives that resolve to black?


- The definition of the tokens in `colors.json` is a bit confusing. SRGB is defined in 0-255, but the `components` array defines values from 0-1. Then, there is a hex code next to that that doesn't get used to generate anything.

When I tried testing changing a color during testing, I changed the hex to try and see that the color went through the build, but it did not. I then investigated and learned that the hex wasn't being used, and now mismatched the components entries. I then found a hex that I wanted to change to, and had to convert that hex to rgb, and then divide each value individually by 255.

That feels quite counterintuitive. I don't feel like we should have two points of definition for a color when the intent of the tokens is to have only one authoritative source, and it feels weird to have to do those calculations for getting to a specific color.

Maybe I am missing some context here though, so that is something I want to be aware of.

pyrello:
> * Then, there is a hex code next to that that doesn't get used to generate anything.

The hex value is included for readability. You may not have been present when I covered this, but the hex value is ~not part of~ only available as a fallback in the DTCG spec. Style Dictionary, the tool we are using to transform the token definitions into CSS variables, uses the sRGB values instead of `hex`.

> What is the reason for two primitives that resolve to black?

Good question and not sure I have a totally satisfying answer. Black would ideally just be defined as part of the neutral ramp (along with white). But it is also a brand color (black and gold), so defining it separate from gold didn't seem right. I'm open to suggestions on the right approach here, but tend to agree it should only be defined once. @bspeare thoughts?

bspeare:
The initial idea on the two black values was that the neutral ramp version could be dialed back in the future for better dark mode readability.  Open to suggestions on best way to proceed.

pyrello:
> The initial idea on the two black values was that the neutral ramp version could be dialed back in the future for better dark mode readability. Open to suggestions on best way to proceed.

The adjustment I made is to let black be part of the ramp for now and gold is just `color.gold`. They get their branding identity at the semantic level with `color.brand.gold` and `color.brand.black`.

pyrello:
> README links the DTCG spec but doesn't say which bits we actually use. Could we get a short section on that? Or an example token file with comments above each one?

We did not include more information in the README about what is used because it is readable in the files themselves and documenting it in the README becomes a drift surface. Can you provide more information about what exactly you are hoping to see? Is there a reason that the token files themselves can't serve as reference here? Or are you just looking for a quick DTCG explainer that doesn't require visiting the DTCG's obtuse docs?

pyrello:
> * The definition of the tokens in `colors.json` is a bit confusing. SRGB is defined in 0-255, but the `components` array defines values from 0-1. Then, there is a hex code next to that that doesn't get used to generate anything.

The specification that we are following does not allow us latitude for deciding how we want to represent our colors: https://www.designtokens.org/tr/2025.10/color/#srgb

CSS can use both the 0-255 and 0-1 scales for sRGB, so either is valid there.

pyrello:
> That feels quite counterintuitive. I don't feel like we should have two points of definition for a color when the intent of the tokens is to have only one authoritative source, and it feels weird to have to do those calculations for getting to a specific color.

I updated my [previous comment](https://github.com/uiowa/uids/pull/1088#issuecomment-5733588981) to reflect the fact that `hex` is part of the DTCG spec, but is only used as a fallback. The sRGB entry is required. So, our options are to leave `hex` in for enhanced readability and deal with the fact that its duplicative or to remove it. I've added a test to ensure that the `hex` and `srgb` values match, so we'll know if they've drifted.

bspeare:
> > README links the DTCG spec but doesn't say which bits we actually use. Could we get a short section on that? Or an example token file with comments above each one?
> 
> We did not include more information in the README about what is used because it is readable in the files themselves and documenting it in the README becomes a drift surface. Can you provide more information about what exactly you are hoping to see? Is there a reason that the token files themselves can't serve as reference here? Or are you just looking for a quick DTCG explainer that doesn't require visiting the DTCG's obtuse docs?

Yup, just a quick DTCG explainer that doesn't require visiting the DTCG's obtuse docs.

GaryRidgway:
> > * The definition of the tokens in `colors.json` is a bit confusing. SRGB is defined in 0-255, but the `components` array defines values from 0-1. Then, there is a hex code next to that that doesn't get used to generate anything.
> 
> The specification that we are following does not allow us latitude for deciding how we want to represent our colors: https://www.designtokens.org/tr/2025.10/color/#srgb
> 
> CSS can use both the 0-255 and 0-1 scales for sRGB, so either is valid there.

I dug into that a bit more too and it makes sense now from the standpoint of how its implemented, so I think that makes more sense now. Thank you for the investment!

GaryRidgway:
> > That feels quite counterintuitive. I don't feel like we should have two points of definition for a color when the intent of the tokens is to have only one authoritative source, and it feels weird to have to do those calculations for getting to a specific color.
> 
> I updated my [previous comment](https://github.com/uiowa/uids/pull/1088#issuecomment-5733588981) to reflect the fact that `hex` is part of the DTCG spec, but is only used as a fallback. The sRGB entry is required. So, our options are to leave `hex` in for enhanced readability and deal with the fact that its duplicative or to remove it. I've added a test to ensure that the `hex` and `srgb` values match, so we'll know if they've drifted.

Ohhhh I like this test, yes, thank you!

GaryRidgway:
> > The initial idea on the two black values was that the neutral ramp version could be dialed back in the future for better dark mode readability. Open to suggestions on best way to proceed.
> 
> The adjustment I made is to let black be part of the ramp for now and gold is just `color.gold`. They get their branding identity at the semantic level with `color.brand.gold` and `color.brand.black`.

I think thats fine and fits well into the intent of the semantic tokens. If we need to lower the brightness for darkmode too, we can always adjust what the semantic tokens point at?

pyrello:
> > > The initial idea on the two black values was that the neutral ramp version could be dialed back in the future for better dark mode readability. Open to suggestions on best way to proceed.
> > 
> > 
> > The adjustment I made is to let black be part of the ramp for now and gold is just `color.gold`. They get their branding identity at the semantic level with `color.brand.gold` and `color.brand.black`.
> 
> I think thats fine and fits well into the intent of the semantic tokens. If we need to lower the brightness for darkmode too, we can always adjust what the semantic tokens point at?

Yes, my understanding is that definitions for dark mode would exist at the semantic layer.

GaryRidgway:
Finally got back around to checking this out again, and it looks good to go!

GaryRidgway (approved):

